PR #37, opened 2015-05-25
Upgrade to conform to M12

cy6erGn0m:
- replace trait with interface
- use size() and length() instead of length
- use split(Regex)
- use arrayOf() instead of array()
- upgrade companion object syntax
- upgrade initializer syntax (init keyword)


hhariri:
Thanks.

hastebrot:
>upgrade initializer syntax (init keyword)

Ohh. I really liked the double curly braces `{{`.

hhariri:
@hastebrot the issue is that Kotlin now warns if you don't have the init keyword and that's one of the reasons we're looking at trying a different approach.

PR #48, opened 2015-09-30
spek-41 init keyword is clunky

paslavsky:
I added constructor to the ``Spek`` class with optional parameter that allow to create tests with the following syntax:
```
class SomeTest: Spek({
    give("...") {
        ...
    }
})
```
This parameter is optional so the backward compatibility is stored


sad2project:
Clever

paslavsky:
Closed due to the conflicts
New PR is https://github.com/JetBrains/spek/pull/71

PR #53, opened 2015-12-12
Removed maven plugin

jitpack-io:
Hi

Since maven-publish plugin is already used having the maven plugin is not necessary. We had some users trying to install spek from JitPack.io and this small change makes it work.

Happy building!


hhariri:
Thanks

jitpack-io:
Build was successful https://jitpack.io/#JetBrains/spek/149f0d9bbe :+1:

PR #56, opened 2016-01-04
Added older and current versions of spek. (Update download.md)

IRus:
Hi, i found that version on download page outdated, so i add historical versions (one spek version per kotlin version) and current spek version.


hhariri:
Thanks!

PR #58, opened 2016-01-29
Removed superfluous code bits

sad2project:
Removed all the qualified names in api.Spek, since they share the same
package and make it a lot harder to read.
Removed all the "public" modifiers, since they just cluttered the code
and were inconsistent.


hhariri:
Thanks for this. This was back from early Kotlin days :)

sad2project:
I thought it might be. And I'm guessing that the qualified names were from when you moved the api to a different package?

hhariri:
Yeah that was actually more like a bug that I didn't notice :)

Issue #39, opened 2015-06-26
afterOn executing before On


trharoldsen:
Expected: on runs before afterOn
Actual: afterOn runs before on

Sample:
```
given("") {
    beforeOn { print("1") }
        on("") {
            print("2")
            it("") {
                print("3")
            }
    }
    afterOn { print("4") }
}
```

output:
1423

hhariri:
Yep it's a known issue. With the restructuring hopefully this will be solved.

jskierbi:
Anything new in this respect?

I understand that ```beforeOn``` and ```afterOn``` are called during adding test task to the queue, with queue itself being processed afterwards. Is there anything I can be helpful with? I really like Spek and want to add it to my project badly, but this issue prevents me from doing so

hhariri:
Is there any possibility of working around it by just calling the code you need yourself in a function call explicitly? Currently because of how Spek is designed, without major changes it's not possible :(

jskierbi:
Hmm I forked repo and started some code rework in this regard, now I've managed to change structure a bit so that given, on and it calls can be deferred after listing them (in master this is done via ```iterateOn```). One problem yet to overcome -> JUnits ParentRunner need to list all children tree before execution, which is unfortunately not possible before running given/on/it expressions (registers children), and this means all expressions are called prior to test execution... Will dig a bit more into how JUnit ParentRunner works and hopefully find a solution :)

Please tell me, are those major changes planned? I don't want to overlap the code - in hope that my changes can be later accepted as a PR :) (I'll ut them ofc!)

hastebrot:
Hi Jakub,

I experimented a bit with ParentRunner a few months ago. Register new
children during test execution works fine.

My plan was to use graph transformations to rewrite Spek classes to graphs
that resemble the execution order and use these graphs for the ParentRunner
tree and test run output. This approach could help to understand and debug
the execution order better and allow custom DSLs, e.g. a Mocha.js-like
hierarchical spec definition.

One problem is, that you have to run parts of the tests in order to get
parts of the spec. The graph transformation approach would traverse the
spec, run a part of the specification, add new nodes to the graph,
transform it to be in execution order and resume the execution.

Maybe this approach (graph transformations) is a bit overkill. AFAIK other
test frameworks don't use this approach. But it is quite popular in the
compiler-world, I saw Oracle's GraalVM (which will be part of JDK 9) uses
it.

hhariri:
[...]

PR #59, opened 2016-01-30
Execution model improvements

jskierbi:
Multiple changes mainly discussed in #39

1. Fixed execution order of ```beforeOn``` and ```afterOn```
2. Added ```beforeEach``` and ```afterEach``` idiom on ```Spek``` level
3. Fixed and improved runners, both Console runner and JUnitClassRunner
4. Improved spek-tests module: now each test is run against both Console runner an JUnitClassRunner, so we can expect that both **runners behaves in exactly same way**
5. Added some new tests in spek-tests module (exception reporting, before/after each)

Remarks:
IntelliJ integration works ok (i.e. JUnit execution tree updates as it should), however because of how integration works, I had to write test execution in unexpected place. Please note ```JUnitClassRunner#getDescription``` and ```JUnitClassRunner#run(RunNotifier)``` - actual tests are executed in first method storing results inside runner, reporting the results in latter method.


hhariri:
Thanks @jskierbi. Assigning to @akozlova

jskierbi:
@hhariri @akozlova please tell me have you found time to look at this pr?

akozlova:
sorry, didn't have time. will check this week.

TwistedEquations:
Any progress on this? It is an ideal solution to an issue im having.

hhariri:
@jskierbi you seen https://github.com/JetBrains/spek/pull/69 ?

jskierbi:
@hhariri yes already commented

Is resolving conflicts for this PR sensible? I use the code already in production for an Android project, but I really don't know whether this is expected dev direction for spek (especially with this hacked JUnit runner, that may not work somewhere...)

akozlova:
Sorry that it took so long for me to process. I was trying to check all this staff against junit5 (lambda), As far as I can see, you'll need to provide separate TestEngine and then you'll be able to provide navigation to the file:line; rerun would be possible because test ids would be unique and serializable so IDEA would be able to leave only selected nodes, etc. So for the time being: what is done is a hack but it should work before junit 5 release. For junit 5, it would be better to provide separate support (need to reconsider in couple of months). Thanks

hhariri:
@jskierbi Is there an email to reach you on? Mine is hadi at jetbrains....

hhariri:
Most of these changes have been incorporated or are now incompatible with the new 1.0 mode. Thanks for all the work, ideas and suggestions.

PR #64, opened 2016-02-24
Removed outdated dependency examples.

tlahn:
The old versions have been removed from the page, but the dependency examples were still there.

### DIFF
--- a/download.md
+++ b/download.md
@@ -29,45 +29,6 @@ Spek has not been released yet and as such is not made available on Maven Centra
 
 To use it in Maven insert the following in your pom.xml file:
 
-
-{% highlight xml %}
- 
- <dependency>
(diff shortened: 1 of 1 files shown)


hhariri:
But those repositories aren't pointing to Maven Central....

tlahn:
In commit 9266ed3d5b025ab25acebb4349e7b0f55ba65822 the versions 0.1.142 - 0.1.188 and the headlines above the dependency examples (lines 39 - 42, 83 - 85) have been removed. But the dependency example for "org.spek:spek" which is for version < 0.1.127 wasn't. This results in duplicate examples, one with "org.spek:spek" and one with "org.jetbrains.spek", without any hint when to use which. The first one doesn't work for current spek versions (as it is the < 0.1.127 example).

PR #70, opened 2016-03-05
Fixed a couple of wrong groupIds

soywiz:
Download page have a couple of groupId `org.spek` that should be `org.jetbrains.spek`.


hhariri:
Thanks!

PR #71, opened 2016-03-06
spek-41 init keyword is clunky

paslavsky:
I added constructor to the ``Spek`` class with optional parameter that allow to create tests with the following syntax:
```
class SomeTest: Spek({
    give("...") {
        ...
    }
})
```
This parameter is optional so the backward compatibility is stored


hhariri:
We've substantially modified the structure and fixed some of the things moving forward. The changes will be in the 1.0 branch which will be for the release.

PR #73, opened 2016-03-29
Improve documentation

kmizu:
* How to add the dependency to Spek (in gradle)
* Maven version should be written in the future

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -55,6 +55,28 @@ Each specification consists of:
 * **on**: One or more of these. Each on is an action. The act. More than one on indicates that they share the same context.
 * **it**: One or more of these. Each it is a consequence of the action. The assert. Normally it's good practice to have a single assertion per it.
 
+#### Dependency
+
+Adding the dependency to Spek is needed to use Spek.
+
+For example, gradle users add the following lines to your `build.gradle`
+
+```groovy
+buildscript {
+  // ...
+  ext.spek_version = '1.0.25' //replace the value of spek_version with your Spek version
+  // ...
+}
+repositories {
+  maven {
+    url 'http://repository.jetbrains.com/all'
+  }
+}
+dependencies {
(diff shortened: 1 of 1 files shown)


kmizu:
How about your concerns about this pull request ?

docs.md
```diff
@@ -55,6 +55,23 @@ Each specification consists of:
...
+#### Dependency
+
+Adding the dependency to Spek is needed to use Spek.
+
+For example, gradle users add the following lines to your `build.gradle`
+
+```groovy
+repositories {
+  maven {
+    url "http://repository.jetbrains.com/all"
```
artem-zinnatullin: nit: single quoutes

docs.md
```diff
@@ -55,6 +55,23 @@ Each specification consists of:
...
+  }
+}
+dependencies {
+  testCompile "org.jetbrains.spek:spek:$your_spek_version" //replace $your_spek_version with correct Spek version (e.g. 0.1.195)
```
artem-zinnatullin: Latest is `1.0.25`, let's insert it right here?

kmizu: OK.

docs.md
kmizu: Although double quotes are also OK,  I'll fix it.

docs.md
hhariri: Yeah but it changes every so often so does it really make sense?

kmizu:
I reflect reviewer's comment and rebased this branch.

docs.md
artem-zinnatullin: Well, repo doesn't have tags so the only way to check if you're on latest
version is to open maven repo and find it out which is not very convenient.

hhariri:
We've addressed this in the new docs. (Maven still pending). Thanks.

PR #75, opened 2016-04-02
Add focused tests

lkogler:
This pull request adds the ability to run only one test or describe block at a time using f-prefixed methods, i.e. `fit` and `fdescribe`.

Structurally, this was implemented by deconstructing the tree into a list of paths, and only running paths that contain at least one focused node.

Although it is a fairly substantial structural change, there shouldn't be too many user-facing side effects, except that the output of the console is different, due to the flattening of the tree structure.  I changed the output to be more concise (only printing a `.` for passing tests), but could add a verbosity flag to the command line options in the future.

One other note: if there is a focused test in the suite, the console runner will only run that test out of the whole suite.  The Junit runner scopes the effect of the focus to a single class, since it is easy to only run one class from IntelliJ using ctrl+shift+R.


hhariri:
I've not used fdescribe at all in Jasmine, and while this in principal doesn't affect those that don't use it, I do wonder if this is actually practical in the sense of having to modify the code each time to focus on specific tests. I realise right now the limitation we have in Spek for running specific ones, but wouldn't an approach such as Categories or something else work in this case?

lkogler:
I'm curious how you would envision "Categories" to work? 

I have used the test-focusing feature quite a bit in Jasmine and Cedar. It's not as nice as just being able to use ctrl+shift+R from the IDE to run one test, but it gets the job done.  It's definitely an important feature to be able to run only one test at a time.

hhariri:
If this is specifically to overcome the issue of running one or two tests as they fail, then yes, it's probably the best workaround possible. But if it's group of tests you'd want deactivated/activated then I guess better categories fit. Nonetheless we can merge this.

PR #100, opened 2016-06-22
Enables resource autoCleanup

SalomonBrys:
This PR does 2 things:
- It makes sure that cleanup code (including `afterEach`) is called even if the test has failed
- It introduces a new idiom for post-run resource cleanup

### Cleanup code is called even if the test has failed

In the current implementation, the `afterEach` callbacks are called after the test **only if it succeeds**, which is, to me, a bug. So this PR corrects that.

### New idiom for post resource cleanup

first, a demo ;)

```kotlin
    it("should demonstrate autoCleanup") {
        val connection = setUpDatabase().autoCleanup()
        /* ... test the database ... */
    }
```

The important part here is the `autoCleanup` call.  
What it does is simple: it registers a callback that will be called once the test has finished, *whether or not it has succeeded* and, in that callback, close the connection.

`autoCleanup` is very usefull for `Closeable` objects, but for any other kind of cleanup, you can register your own callbacks:

```kotlin
    it("should demonstrate onCleanup") {
        val network = setUpNetwork()
        onCleanup {
            /* ... complicated cleanup ... */
        }
        /* ... test the network ... */
    }
```

Note that `beforeEach` callbacks are also able to register `onCleanup` callbacks and `autoCleanup` resources, which brings me to the...

### C-C-C-C-Combo!

To my opinion, this PR should deprecate `afterEach`. Let me explain.

Imagine a Spek class hierarchy. For example, we have a main class that sets up a database, beaucause all tests of the project will need it, and a sub-class for a test sub-set that will need a network connection.  
Now, each class registers its own resource on `beforeEach` and cleanup in `afterEach`:

```kotlin
    beforeEach {
        this.resource = openResource()
    }
    afterEach {
        this.resource.close()
    }
```

Feel free to replace `resource` by `database` or `network`. The issue is simple. There is no way to link resource opening to the resource cleanup. Because `afterEach` is called every time, this can cause trouble if one resource has been opened and the second throws an exception upon opening. Only one resource will be opened, but both closing code will be called, which can lead to unpleasant situations.  
To correct this, we have to check whether the resource was opened:

```kotlin
    beforeEach {
        val opened = true
        this.resource = openResource()
    }
    afterEach {
        if (opened)
            this.resource.close()
        opened = false
    }
```

Yikes.

What this PR introduces is, IMHO, better and easier:

```kotlin
    beforeEach {
        this.resource = openResource()
        onCleanup {
            this.resource.close()
        }
    }
```

Or better yet:

```kotlin
    beforeEach {
        this.resource = openResource().autoCleanup()
    }
```

This idiom makes sure that the `onCleanup` call is called *only if* the `beforeEach` callback was called, effectively linking the resource opening and cleanup.

That's why I think `afterEach` should be deprecated. It's better to link an `onCleanup` code to the success of a `beforeEach` than registering a `afterEach` that may be called "orphaned".

### DIFF
--- a/spek-core/src/main/kotlin/org/jetbrains/spek/api/AssertionBody.kt
+++ b/spek-core/src/main/kotlin/org/jetbrains/spek/api/AssertionBody.kt
@@ -1,0 +1,14 @@
+package org.jetbrains.spek.api
+
+import java.io.Closeable
+
+interface AssertionBody {
+
+    fun onCleanup(block: () -> Unit)
+
+    fun <T : Closeable?> T.autoCleanup(): T = apply { if (this != null) onCleanup { close() } }
+
+// When JDK 7 is officially supported:
+//    fun <T : AutoCloseable?> T.autoCleanup(): T = apply { if (this != null) onCleanup { close() } }
+
+}
--- a/spek-core/src/main/kotlin/org/jetbrains/spek/api/DescribeBody.kt
+++ b/spek-core/src/main/kotlin/org/jetbrains/spek/api/DescribeBody.kt
@@ -5,11 +5,11 @@ interface DescribeBody  {
     fun xdescribe(description: String, evaluateBody: DescribeBody.() -> Unit)
     fun fdescribe(description: String, evaluateBody: DescribeBody.() -> Unit)
 
-    fun it(description: String, assertions: () -> Unit)
-    fun xit(description: String, assertions: () -> Unit)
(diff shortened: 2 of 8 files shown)


SalomonBrys:
I've proposed the exact same thing for the branch junit5 in #101. However, I still think this PR is valid for the current version of Spek ;)

raniejade:
I appreciate the PR, however we won't be doing any maintenance on the current version as it will be eventually replaced.

PR #101, opened 2016-06-23
Enables resource autoCleanup for Junit5

SalomonBrys:
This PR is the same as #100 but for the branch junit5. You can read its description for a full understanding of this PR ;)

The exact same description applies:
- It makes sure that cleanup code (including `afterEach`) is called even if the test has failed
- It introduces a new idiom for post-run resource cleanup (`onCleanup` and `Closeable.autoCleanup`)

### DIFF
--- a/spek-api/src/main/kotlin/org/jetbrains/spek/api/dsl/Dsl.kt
+++ b/spek-api/src/main/kotlin/org/jetbrains/spek/api/dsl/Dsl.kt
@@ -6,9 +6,9 @@ package org.jetbrains.spek.api.dsl
  */
 interface Dsl {
     fun group(description: String, pending: Pending = Pending.No, body: org.jetbrains.spek.api.dsl.Dsl.() -> Unit)
-    fun test(description: String, pending: Pending = Pending.No, body: () -> Unit)
+    fun test(description: String, pending: Pending = Pending.No, body: TestBody.() -> Unit)
 
-    fun beforeEach(callback: () -> Unit)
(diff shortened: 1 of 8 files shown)


raniejade:
Not convinced if this the right DSL to go with, see #102.

SalomonBrys:
Well, first of all, this PR makes `afterEach` closures called whatever the test result. Independently of the DSL, I think this issue needs to be addressed as this is, IMHO, a major issue (or maybe it's just me misunderstanding the semantic of `afterEach`).

----

Regarding the DSL, I feel I should expose the goals of what I'm trying to achive here:

- A DSL that enables clean-up callbacks on a *per test* basis. Per group is also great, but was not my primary goal.
- A DSL that enables `beforeEach` closures to register their counterpart. As stated in #100, I don't think `afterEach` is a good idea. In fact, I think it should be deprecated. Each cleanup closure that was registered before a throw should be called, and those only. So if there's a throw in a `beforeEach`, any cleanup closure registered by previous `beforeEach` will be called.
- A way to automagically close Closables opened in either `beforeEach` or in the test itself, with the least disruptive syntax possible (hence the simple `.autoCleanup()`)

I'm not emotionally attached to the DSL itself and I think the `memoized` idea could be great, provided that it meets the above criteria ;)

raniejade:
Yeah, `afterEach` should be invoked regardless of the test result - I'll verify this. `memoized` will meet every criteria, plus there won't be no need for the extra call to `autoCleanup` in a `beforeEach`.

PR #103, opened 2016-06-30
Update docs.md

jasperblues:
- Most users will be familiar with BDD-style frameworks and terminology. 
- Remove repeated assertion. Once is enough.


hhariri:
We've fixed this manually. Thanks.

jasperblues:
Looks nice.

PR #112, opened 2016-08-31
Bad dependency name

fboldog:
spek-junit-platform-engine not exsists only spek-junit-platform-engine
http://search.maven.org/#artifactdetails%7Corg.jetbrains.spek%7Cspek-junit-platform-engine%7C1.0.77%7Cjar


hhariri:
Thanks!

PR #121, opened 2016-10-02
Add return statement when artifact related values are "nothing".

hotchemi:
## Change

- Add return statement when artifact related values are "nothing".
- This change is just in case for contributor's local env.

### Why?

When I tried to build this on my local, I got the following error since it failed to authenticate.

screen shot 2016-10-02 at 14 45 11

### DIFF
--- a/spek-dist/build.gradle
+++ b/spek-dist/build.gradle
@@ -142,6 +142,9 @@ bintray {
 }
 
 artifactory {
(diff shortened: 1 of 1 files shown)


hhariri:
Thanks. I think we're actually going to end up removing artifactory anyway, but will merge for now.

PR #146, opened 2017-01-01
Add table driven tests capabilities

### LINKED ISSUE #91: Support table driven tests
Greetings!

Are there any plans to support table driven tests like the ones implemented in Spock using [where](https://spockframework.github.io/spock/docs/1.0/data_driven_testing.html) or the JUnit [Parameterized](https://github.com/junit-team/junit4/wiki/Parameterized-tests) tests?

Thanks!

FaustXVI:
Closes #91 

Todo :
- [X] Add arrity until 2
- [x] Validate API
- [x] Find a way to support any arrity or define what the maximum arrity should be


FaustXVI:
Here is how I think it could be done :

 ```kotlin
describe("subtract") {
        unroll(
                testCase(4, 2),
                testCase(5, 3)
        ) { a, b ->
            it("should return the result of subtracting $b from $a") {
                val subtract = Calculator().subtract(a, b)
                assertEquals(2, subtract)
            }
        }
    }
```

This code is actually working on my branch.
Any idea on how to improve that API ?

**Comment by FaustXVI** [comment]
I have done it for all arity kotlin allows us (more and it can't find Function23).
I wrote some code to generate the code and the test, I don't know if it's a good idea to add it to the sources. What do you think ?

**Comment by raniejade** [comment]
Can you also play around with this:

```kotlin
given { param1, param2 ->
    it("should do something with $param1 and $param2") {
        ...
    }
} where {
    testCase(1, 2) // row(1, 2)
}
```

Also if `TestCase` is a `data class`, in Kotlin `1.1` we could use destructuring lambdas.

FaustXVI:
Here is what I get :
```
                  given { p1: Int ->
                        it("should be visible with ints $p1") {}
                    }.where(
                            testCase(1),
                            testCase(2)
                    )
```

I find some drawbacks to this :
- I can't get where as a infix function because it use a variable arity as second parameter
- I *must* declare the type of the parameters in the lambda because the type inference fails

What do you think ?

raniejade:
instead of accepting an array of `testCase`, `where` could accept a lambda. You call `testCase` inside the lambda. I'm waiting for others to chip-in @hhariri :)

hhariri:
So one thing I had initially in spek was givenData, which was basically like test cases, but I did wonder what real benefit it provided ever people just having the flexibility of defining their data and passing it in as they'd want. 

In essence, I'm not against the idea of having test cases, but I think whatever syntax we provide should have some substantial benefit, whether it's clarity, discoverability, ability to do reporting on it, etc.

raniejade:
> So one thing I had initially in spek was givenData, which was basically like test cases, but I did wonder what real benefit it provided ever people just having the flexibility of defining their data and passing it in as they'd want.
>
 > In essence, I'm not against the idea of having test cases, but I think whatever syntax we provide should have some substantial benefit, whether it's clarity, discoverability, ability to do reporting on it, etc.

Makes sense, flexibility is a big plus since we can't please everyone :). @geond @russel @sksamuel

FaustXVI:
To me, the goal of table driven tests (or parametrized tests) is just to avoid duplication in tests and make it more readable. For example, on the fizzBuzz kata, we can avoid writing tests like
```kotlin
it("should buzz for 3")
it("should buzz for 6")
it("should buzz for 9")
```
and just have 
```kotlin
it("should buzz for multiples of 3")
```
which is a spec of higher abstraction and so way closer to the business.
Finally, this is just the first step to property based testing which allow even higher abstractions.

Writing this, I'm thinking of another possible syntax : 
```kotlin
it("should buzz for multiples of 3") { parameter ->
  assertionHere
} withExemples {
  exemple(3)
  exemple(6)
  exemple(9)
}
```

FaustXVI:
Another syntax could also be : 
```kotlin
    with {
        exemple(3)
        exemple(6)
        exemple(9)
    } it "should buzz for multiple of 3" {

    }
```

hhariri:
I completely agree regarding keeping it DRY (albeit in certain cases DRY I think shouldn't apply in tests, but not in this particular one). However, how is that different from doing something like a for loop?

FaustXVI:
Well, table driven tests are very similar to a for loop, but it's main goal is to remove noise from the language used by proposing a DSL. Let's do the fizzBuzz sample with loops. We can do it with many different ways : 

I'll list here the ways I think we can use for loops for that but I'm still (very) new to kotlin so I may miss something. If so, please let me know.

First let's use for loop in the test itself using ranges
```kotlin
it("should buzz for multiples of 3") {
    for (i in 1..3) {
        fizzBuzzOf(i * 3) shouldEqual "buzz"
    }
}
```
This does the trick but if it fail, you won't know which case fails 😢 so we can move the for out of the test : 
```kotlin
for (i in 1..3) {
    it("should buzz for multiples of 3 (case ${i * 3})") {
        fizzBuzzOf(i * 3) shouldEqual "buzz"
    }
}
```
That's better 😃 but we need to multiply it by 3 to get the value we want 😢 so we can use lists for that : 
```kotlin
for (i in listOf(3,6,9)) {
    it("should buzz for multiples of 3 (case $i)") {
        fizzBuzzOf(i) shouldEqual "buzz"
    }
}
```
Ok ! Know that's fair enough for me in this specific case ! Let's now imagine that we want to test the function `add` of a class `Complex` that represent Complex numbers. We then need two parameters. If we use the same strategy, we get : 
```kotlin
for (complexes in listOf(
listOf(Complex(1, 2), Complex(3, 4)),
listOf(Complex(5, 6), Complex(7, 8))
)) {
    val complex1 = complexes[0]
    val complex2 = complexes[1]
    it("should add two complexes (case $complexes)") {
        complex1 + complex2 shouldBe Complex(complex1.real + complex1.real, complex1.imaginary + complex2.imaginary)
    }
}
```
we start to get some boiler plate code here : 
- `for (complexes in listOf(` is always used and is getting us away of the purpose of the test
- `val complex1 = complexes[0]` has the same drawback (and we can't use lambda destructuring since this is a list)
- `listOf(Complex(1, 2), Complex(3, 4))` the use of the word `listOf` is not domain specific to tests
- `(case $complexes)` doesn't add any useful information to my spec except when it's failing

For me, all of this keep me away from having an easy to read specification. If we use one of the syntax proposed before, we'll get : 
```kotlin
with {
    example(Complex(1, 2), Complex(3, 4))
    example(Complex(5, 6), Complex(7, 8))
} it "should add two complexes" { complex1, complex2 ->
    complex1 + complex2 shouldBe Complex(complex1.real + complex1.real, complex1.imaginary + complex2.imaginary)
}
```
- the words `with` associated with `example` is more test specific than `for (complexes in listOf(` and `listOf`
- destructuring of the example is hidden within the framework (but could be delegated to the user)
- the string `(case $complexes)` can be added by the framework

That was a long post :) but at least I think it explains why I like parametrized tests 😸

raniejade:
> I completely agree regarding keeping it DRY (albeit in certain cases DRY I think shouldn't apply in tests, but not in this particular one). However, how is that different from doing something like a for loop?

For simple cases it will be the same, however readability will be impacted if case gets too complex.

raniejade:
I still like:

```kotlin
given { param1: Int, param2: String -> 
    ...
} where {
   testCase(1, "foo")
   testCase(2, "bar")
}
```

The extra parameter type hints gives value to readability.

hhariri:
That's a good summary. Thanks.

FaustXVI:
@raniejade I do agree that type hints may add some readability but enforcing them seems out of kotlin's philosophy to me. It's a battle against the type inference and may not be very welcome by users since it probably is a feature they do like.
I think letting the user choose to add type hints or not allows us to satisfy what ever style suits the user.
@hhariri any preference ?

hhariri:
@FaustXVI I agree in that we should keep type hinting optional.

FaustXVI:
I played a bit with that and, in order to keep type hinting optional, I need to use a vararg argument for the examples at the beginning like : 
```kotlin
                    with(
                            testCase(1),
                            testCase(2)
                    ) it "should be greater than 0" { p1 ->
                        p1 shouldMatch greaterThan(0)
                    }
```
I also tried to allow the user to use the parameter in the name of the test and get : 
```kotlin
                    with(
                            testCase(1)
                    ) { p1 ->
                        it("$p1 should be greater than 0") {
                                p1 shouldMatch greaterThan(0)
                        }
                    }
```

What do you think ?

hhariri:
Pending trying it out, at initial glance it looks viable. @raniejade ?

raniejade:
Looks good to me. Up next is the implementation :)

FaustXVI:
Here it is :)

settings.gradle
```diff
@@ -7,3 +7,4 @@ include 'spek-junit-platform-engine'
 include 'spek-dist'
 include 'documentation'
 include 'spek-samples'
+include 'spek-table-driven-extension'
```
raniejade: `spek-data-driven-extension`? @hhariri

spek-samples/src/test/kotlin/org/jetbrains/spek/samples/subject/CalculatorTableSpec.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.jetbrains.spek.api.dsl.describe
+import org.jetbrains.spek.api.dsl.it
+import org.jetbrains.spek.table.example
+import org.jetbrains.spek.table.with
+import kotlin.test.assertEquals
+
+object CalculatorTableSpec : Spek({
+    describe("subtract") {
+        with(
+                example(4, 2),
```
raniejade: Can you fix the indent :) it should always be 4 spaces

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
```diff
@@ -0,0 +1,320 @@
...
+import org.jetbrains.spek.api.dsl.Spec
+import org.jetbrains.spek.api.dsl.TestBody
+import org.jetbrains.spek.api.dsl.TestContainer
+
+data class Example1<out P1> internal constructor(val p1: P1)
+
+fun <P1> example(p1: P1): Example1<P1> = Example1(p1)
+fun <P1> Spec.with(vararg examples: Example1<P1>, testContent: TestContainer. (P1) -> Unit) {
+    examples.forEach {
+        val proxyContainer = object : TestContainer by this@with {
```
raniejade: Should we allow nesting of groups? Currently it only allow `test` scopes.

```kotlin
with(
    example(1, 2)
) { p1, p2 -> 
    on("this") {
       it("should do something") { ... }
    }
}
```

@hhariri thoughts?

raniejade (changes requested):

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
hhariri: @raniejade TBH I think it should be consistent with what we already provide.

FaustXVI: I'm sorry, I don't know what that mean. Should we allow nesting ?

raniejade: You should allow other scopes as well, not only `it`s. Quick fix is to use `SpecBody` instead of `TestContainer` :)

FaustXVI: Thanks !
Looking at the SpecBody, I see the `memoized` and  `before/afterXXX` functions. I can't see a use case to that with table driven tests, do you ?
If not, I think introducing a new interface that contains only group and action would be nice. What do you think ?

FaustXVI: Trying to implement that, I don't know where to add the information of the current case.
The only doable thing right now it to add it at the top level. It seems fair enough to me. What about you ?

FaustXVI: @raniejade @hhariri any thoughts ?

raniejade: Sorry @FaustXVI forgot to reply. I take back what I said before, I think it should just contain `test` scopes. As you said, it will be hard to find a place to put the information of the current case.

FaustXVI: Ok, so, any change to do to this PR ?
`spek-data-driven-extension` maybe ? @hhariri ?

raniejade: Oh yeah, can you add `@Experimental` to the public APIs you have added? See `spek-subject-extension` for an example.

raniejade: Documentation would be nice too :)

FaustXVI:
Hi, I just made the requested changes but doing so, I realize that my code is not working within actions.
A code like : 
```kotlin
    describe("a calculator") {
        val calculator = Calculator()
        on("addition") {
            with(
                example(2, 4),
                example(3, 3)
            ) { a, b ->
                it("should return the result of adding two numbers together") {
                    assertEquals(6, calculator.add(a, b))
                }
            }
        }
    }
```
Just doesn't works ! It complies but do not run the test ! I must admit I just do not how it can compile since the action body only calls to the test method… :/
Any idea of what's going on ?

raniejade:
I think the issue is in scoping.

`on` changes the scope, so **this** will be an `ActionBody`. `with` is declared as an extension body of `Spec`. So the call to with using the scope of the outer group which is `describe("a calclulator")`. I think `with` should be an extension of `TestContainer` rather than `Spec`.

FaustXVI:
Yep, it does solve the problem so I changed the code in order to fix that. But still, don't you think we should find a way to avoid that kind of problem in the future ? (not in that PR by the way)
Or at least throw an error at runtime ? The current behavior is to do nothing silently.

raniejade:
It will happen when we adopt Kotlin `1.1` https://github.com/Kotlin/KEEP/blob/master/proposals/scope-control-for-implicit-receivers.md.

raniejade:
Give me this week to do the final review and play around with it. If everything is good I'll merge this :smile:

FaustXVI:
😸 pur-fect !

raniejade:
@artem-zinnatullin you guys might be interested with this :)

artem-zinnatullin:
Yup, will review in 24h!

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
```diff
@@ -0,0 +1,384 @@
+package org.jetbrains.spek.table
+
+import org.jetbrains.spek.api.dsl.Pending
+import org.jetbrains.spek.api.dsl.Spec
+import org.jetbrains.spek.api.dsl.TestBody
+import org.jetbrains.spek.api.dsl.TestContainer
+import org.jetbrains.spek.meta.Experimental
+
+data class Example1<out P1> internal constructor(val p1: P1)
```
artem-zinnatullin: These classes also need to be marked as `@Experimental`

artem-zinnatullin (changes requested):

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
```diff
@@ -0,0 +1,384 @@
...
+}
+
+
+data class Example2<out P1, out P2> internal constructor(val p1: P1, val p2: P2)
+
+@Experimental
+fun <P1, P2> example(p1: P1, p2: P2): Example2<P1, P2> = Example2(p1, p2)
+
+@Experimental
+fun <P1, P2> TestContainer.with(vararg examples: Example2<P1, P2>, testContent: TestContainer. (P1, P2) -> Unit) {
```
raniejade: `given`? Like given these test cases... It feels kinda weird tho.

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
FaustXVI: Are you sure ?
The constructor is internal… nobody should rely on

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
FaustXVI: `forEvery` ?

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
raniejade: It's still part of the public api.

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
raniejade: > `forEvery` ?

👍

artem-zinnatullin: :+1:

artem-zinnatullin:
@raniejade can you please hold on for few days on merging this? @igor-korotenko and me are working on a different implementation that we think fits better into current design of Spek.

raniejade:
Please also include @FaustXVI in your discussions.

spek-samples/src/test/kotlin/org/jetbrains/spek/samples/subject/CalculatorTableSpec.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.jetbrains.spek.api.dsl.it
+import org.jetbrains.spek.api.dsl.on
+import org.jetbrains.spek.table.example
+import org.jetbrains.spek.table.with
+import kotlin.test.assertEquals
+
+object CalculatorTableSpec : Spek({
+    describe("subtract") {
+        with(
+            example(4, 2),
```
raniejade: Just use 3 params and assume the 3rd parameter is the expected value `example(a, b, result)`.  In some cases you don't have an expected value to assert but an expected side-effect (exception thrown, etc ...) when certain parameters are used.

```kotlin
forEvery(
    ...
) {
    expect({ some.foo(param1, param2) }).throws<SomeException>()
}
```

artem-zinnatullin: You can set exception as expected *value*, catch actual and compare with expected.

raniejade: Seems less flexible if we add an `expected` param.
```kotlin
forEvery(
   example(param1_value, param2_value, expected = SomeException::class)
) { param1, param2 ->
   it("should throw SomeException") {
       expect({ some.foo(param1, param2) }).throws(expected)
   }
}
```

There will be cases that you only have one expected value/side-effect, it's to verbose if we include the expected value/side-effect for every example.
```kotlin
forEvery(
   example(param1_value, param2_value)
) { param1, param2 ->
   it("should throw SomeException") {
       expect({ some.foo(param1, param2) }).throws<SomeException>()
   }
}
```

FaustXVI: Adding the expected result in the data is an anti-pattern to me so I'd like not to encourage it (but I still want to allow it). Here is why : 
If you code the fizzBuzz kata this way, you'll end up with only one test with everything in it. Like this : 
```kotlin
forEvery(
   example(1, "1"),
   example(2, "2"),
   example(3, "fizz"),
   example(5, "buzz"),
   example(6, "fizz"),
   example(10, "buzz"),
   example(15, "fizzbuzz"),
   example(30, "fizzbuzz")
) { n, expected ->
   it("should work") {
       expect(fizzBuzz(n)) shouldBe expected
   }
}
```
It seems good at first but there is a very bad side effect : your documentation know sucks!
Instead, you can use it way better by fixing something (like the expected result) in your test. You'll get something like : 

```kotlin
forEvery(
   example(1),
   example(2)
) { n ->
   it("should echo the input") {
       expect(fizzBuzz(n)) shouldBe n.toString()
   }
}

forEvery(
   example(3),
   example(6)
   ) { n, expected ->
   it("should fizz") {
       expect(fizzBuzz(n)) shouldBe "fizz"
   }
}
.
.
.
```
Now the documentation is much nicer (and totally in the philosophy of spek).

Never the less, allowing the user to specify input *and* output in the exemple still necessary, because, sometimes, it still is the right way to do !
Let's use the roman numeral kata as on exemple : 
```kotlin

```kotlin
forEvery(
   example(1, "I"),
   example(5, "V"),
   example(10, "X"),
   example(50, "L")
) { n, expected ->
   it("should translate roman literals") {
       expect(toRoman(n)) shouldBe expected
   }
}
```

spek-table-driven-extension/src/main/kotlin/org/jetbrains/spek/table/TableDriven.kt
FaustXVI: I like the `forEvery` version because it will neatly be enhance by property based testing. We'll get something like : 
```kotlin
forEvery(Int.class) { n ->
...
}
```
If we want two different APIs, `givenData` is better but sounds awkward to me :(

spek-samples/src/test/kotlin/org/jetbrains/spek/samples/subject/CalculatorTableSpec.kt
raniejade: :+1: on generators, would love to see that in spek.

FaustXVI:
Oops, I messed up my last commit message… I'll squash all my commits just before merging, are you ok with that ?

raniejade:
@FaustXVI sure :)

artem-zinnatullin:
Damn, sorry for holding you guys, @igor-korotenko if it's not ready yet, maybe just post everything you have at the moment as WIP PR so guys could check it?

igor-korotenko:
Sorry, for holding you, guys. @artem-zinnatullin, I'll try to do it first thing in the morning tomorrow!

igor-korotenko:
PTAL https://github.com/JetBrains/spek/pull/180

FaustXVI:
Should I close this PR since we go for #180?

artem-zinnatullin:
Feels kinda bad that we've opened PR for same feature, sorry for that…  Looks like this one should be closed then.

We'll mention you as huge contributor to this feature and if you have any coming improvements for this feature — feel free to PR once we merge #180! Thank you for you PR!

raniejade:
We won't have #180 without this PR, everyone involved will be credited 😄.

FaustXVI:
Hi,
sorry for the delay :(
Thanks for the credits :) i'll continue to give my feedback of course :)
The goal is not to have my code in spek but to make spek better :)

rusio:
For some reason the documentation on http://spekframework.org/docs/latest/ lacks a section on table-driven tests. Any idea how to get up-to date documentation?

PR #164, opened 2017-01-19
Feature/print stacktrace for exceptions

caipivara:
Related to https://github.com/JetBrains/spek/issues/160
 
If the test code throws an exception inside an action body there are no stacktrace to know what happened.

### DIFF
--- a/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/ThrowableCollector.kt
+++ b/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/ThrowableCollector.kt
@@ -25,10 +25,10 @@ class ThrowableCollector {
 
     fun isEmpty() = throwable == null
 
-
     fun assertEmpty() {
-        if (throwable != null) {
-            throw throwable!!
+        throwable?.let {
+            if (it !is AssertionError) it.printStackTrace()
+            throw it
         }
     }
 }


raniejade:
Looks good, will merge after I do some testing 👍

spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/ThrowableCollector.kt
```diff
@@ -10,7 +10,7 @@ class ThrowableCollector {
         if (this.throwable == null) {
             this.throwable = throwable
         } else {
-            (this.throwable as java.lang.Throwable).addSuppressed(throwable)
+            (this.throwable as? java.lang.Throwable)?.addSuppressed(throwable)
```
artem-zinnatullin: This seems unneded since there is a branch `if (this.throwable == null)` before and all throwables are children of `java.lang.Throwable`

spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/ThrowableCollector.kt
```diff
@@ -25,10 +25,10 @@ class ThrowableCollector {
 
     fun isEmpty() = throwable == null
 
-
     fun assertEmpty() {
-        if (throwable != null) {
-            throw throwable!!
+        throwable?.let {
+            if (it !is AssertionError) it.printStackTrace()
```
artem-zinnatullin: Hm, not sure about printing here, it may lead to double stacktrace in log and confuse users…

caipivara: @artem-zinnatullin but there's another place where the stacktrace is been printed?

artem-zinnatullin: Well if we're `throwing it` here, shouldn't it be printed by more general catch? Or this throw crashes execution?

caipivara: @artem-zinnatullin currently it crashes and says "Empty Test Suite" so there was no way to know that an exception was thrown or what happened. https://github.com/raniejade/spek-idea-plugin/issues/29

artem-zinnatullin: Got it, ok then 👍

caipivara: But now I agree that this may not be the best place to print it

PR #174, opened 2017-01-29
Allow local delegation for LifecycleAware objects

### LINKED ISSUE #172: Allow subjects to be delegated locally
In Kotlin `1.1` subjects can't be delegated locally, there is a compile error.

```kotlin
object MySpec: SubjectSpek<Foo>({
    val subject by subject { ... }
})
```

The issue is caused by how `LifecycleAware` is implemented (this also means that using memoized causes the same issue).

```kotlin
interface LifecycleAware<T>: ReadOnlyProperty<LifecycleAware<T>, T>
```

It should be declared as
```kotlin
interface LifecycleAware<T>: ReadOnlyProperty<Any, T>
```

raniejade:
Resolves #172.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 kotlinVersion=1.0.6
 junitPlatformVersion=1.0.0-M3
-kotlin.incremental=false
+kotlin.incremental=true
 githubRepo=https://github.com/JetBrains/spek
 issueTracker=https://github.com/JetBrains/spek/issues
 artifactory_user=http://nowhere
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -1,6 +1,6 @@
 kotlinVersion=1.0.6
 junitPlatformVersion=1.0.0-M3
-kotlin.incremental=true
+kotlin.incremental=false
```
artem-zinnatullin: Why?

raniejade: Forgot to remove that. Compiler was acting wonky, my default behaviour is to disable incremental compilation :).

artem-zinnatullin: Ah, we had such problems like half year ago, currently incremental compilation is very stable and damn it saves lots of time!

Issue #176, opened 2017-02-02
exception in action results in green test


robstoll:
The error can be reproduced by the following code:

```kotlin
describe("error sceenario"){
    on("exception thrown in action block"){
        throw IllegalArgumentException("oh no...")
        test("on should be red not green"){}
    }
}
```
The `on` should not be green (I am using intellij) but red and an exception should be shown.

robstoll:
Hm... I realised that it also happens if one of the test fail. Am I using `on` wrong?

raniejade:
@robstoll It's a limitation of the plugin, unfortunately (https://github.com/raniejade/spek-idea-plugin/issues/29). If you're using gradle everything should work as expected. I'm looking into ways to fixing this, but it might take a while.

robstoll:
@raniejade you are right, it works fine in gradle, at least my first concern (described in the issue). However, my second concern (described in my comment) still holds. If an unexpected exception occurs in a group scope then none of the tests are executed. Is this intentional?

raniejade:
@robstoll No, it's not. Re-opening the ticket, it should still continue and just ignore the failing group.

robstoll:
Awesome, when is your next release planned?

PR #177, opened 2017-02-02
catch unexpected exceptions in groups

robstoll:
Edited in editor, so no guarantee that it has not mistakes. An idea how to deal with #176


raniejade:
Thanks! I have an idea how to fix this.

PR #180, opened 2017-02-08
[WIP] Data-driven spek support.

igor-korotenko:
As promised in https://github.com/JetBrains/spek/pull/146 take a look at idea of how data-driven specifications can be implemented in Spek. This is mostly a WIP, so comments/ideas are welcome. 
Differences over https://github.com/JetBrains/spek/pull/146:
- Ability to include expected result for data set into the row.
- Another syntax for declaring data table. No new keyword needed, just extension for `on`
- Execution result with input and expected values printed into test report.
screen shot 2017-02-08 at 15 31 50

Things TDB:
- [ ] Find out a nice way how to be with Exceptions check (ideas welcome)
- [x] Add extensions and Data* classes with higher arity (up to 22)


raniejade:
I'm not convinced with `expected`, we lose flexibility.

It's too verbose:
```kotlin
on("foo.bar()",
           data(1, 2, expected = SomeException::class),
           data(3, -2, expected = SomeException::class)
) { param1, param2, expected ->
    it("should throw $expected") {
        assertThat({ 
            foo.bar(param1, param2)
        }, throws(expected))
    }
}
```

Compared to this one:
```kotlin
on("foo.bar()",
           data(1, 2),
           data(3, -2)
) { param1, param2 ->
    it("should throw SomeException") {
        assertThat({ 
            foo.bar(param1, param2)
        }, throws<SomeException>())
    }
}
```

You can still use one of the parameters (preferably the last) as the `expected` value.

```kotlin
on("addition",
           data(1, 2, 3),
           data(1, 4, 5)
) { param1, param2, expected ->
    it("$param1 + $param2 = $expected") {
        assertThat(param1 + param2, equalTo(expected))
    }
}
```

artem-zinnatullin:
>You can still use one of the parameters (preferably the last) as the expected value.

Neat idea, @igor-korotenko what do you think about that? 

As an alternative we can use something like `object NO_EXPECTED_VALUE` as a default value: `expected: E = NO_EXPECTED_VALUE` so user won't have to specify it if it's not required.

raniejade:
> As an alternative we can use something like object NO_EXPECTED_VALUE as a default value: expected: E = NO_EXPECTED_VALUE so user won't have to specify it if it's not required.

The extra baggage is still there tho:

```kotlin
on("foo.bar()",
           data(1, 2),
           data(3, -2)
) { param1, param2, expected ->
    it("should throw SomeException") {
        assertThat({ 
            foo.bar(param1, param2)
        }, throws<SomeException>())
    }
}
```

igor-korotenko:
@raniejade @artem-zinnatullin seems quite ok for me, the only thing that we lose nice `expected=` named argument if we fully get rid of it. Idea of @artem-zinnatullin with default `NO_EXPECTED_VALUE `looks nicer for me because in kotlin 1.1 we can omit unused param like this.
```kotlin
on("foo.bar()",
           data(1, 2),
           data(3, -2)
) { param1, param2, _ ->
    it("should throw SomeException") {
        assertThat({ 
            foo.bar(param1, param2)
        }, throws<SomeException>())
    }
}
```
This way we are more explicitly declaring that expected argument doesn't matter for us.

raniejade:
It feels a bit too verbose for me.

```kotlin
on("addition",
           data(1, 2, expected = 3),
           data(1, 4, expected = 5)
) { param1, param2, expected ->
    it("$param1 + $param2 = $expected") {
        assertThat(param1 + param2, equalTo(expected))
    }
}
```

compared to this

```kotlin
on("addition",
           data(1, 2, 3),
           data(1, 4, 5)
) { param1, param2, expected ->
    it("$param1 + $param2 = $expected") {
        assertThat(param1 + param2, equalTo(expected))
    }
}
```

artem-zinnatullin:
Yeah, the thing I like about Kotlin in general in compare to Groovy or Scala for instance is its explicitness. Would be great to continue keeping Kotlin's balance between conciseness and explicitness in Spek.

Named arguments are optional, so it's totally up to user whenever to specify them or not:

```kotlin
on("addition", data(1, 2, expected = 3), data(1, 4, 5) /* 5 is expected here. */) { 
    param1, param2, expected ->

    it("$param1 + $param2 = $expected") {
        assertThat(param1 + param2, equalTo(expected))
    }
}
```

What I see in our team is that if function is not used very often — we would prefer to use named arguments so it'll be clear what do they mean for everybody on the code review and during working with the code, once function becomes used pretty frequently — we start to omit named arguments since everybody is familiar with their meaning/order. That's the beauty of named arguments.

igor-korotenko:
Kotlin enables us to have win-win variant here, none of them excluding another.
We should remember that explicit named arguments are still optional, also last value will be optional because it will have default value. So this is about flexibility and everybody still can use variant he likes most.

- With `expected` named arg
```kotlin
val data = arrayOf(
    data(1, 2, expected = 3),
    data(1, 4, expected = 5)
)

on("addition",  with = *data) { param1, param2, expected ->
    it("$param1 + $param2 = $expected") {
        assertThat(param1 + param2, equalTo(expected))
    }
}
```

- Without `expected` named arg
```kotlin
val data = arrayOf(
    data(1, 2, 3),
    data(1, 4, 5)
)

on("addition",  with = *data) { param1, param2, expected ->
    it("$param1 + $param2 = $expected") {
        assertThat(param1 + param2, equalTo(expected))
    }
}
```

- Without `expected` value at all
```kotlin
val data = arrayOf(
    data(1, 3),
    data(0, 4)
)

on("addition",  with = *data) { param1, param2, _ ->
    it("$param1 + $param2 = 4") {
        assertThat(param1 + param2, equalTo(4))
    }
}
```

raniejade:
I don't really see the value of having an extra `expected` parameter.  Just by looking at the lambda parameters, I can clearly tell what each parameters mean.

`data(1, 2, 3)` -> `{ param1, param2, expected ->`.

Even if we allow it to be optional (in cases when there is no expected value), you still have `n + 1` lambda parameters.

```
on(...) { param1, param2, _ ->
}
```

igor-korotenko:
n + 1 lambda parameter here

```kotlin
on(...) { param1, param2, _ ->
}
```

is not something redundant, rather a hint to maintainer that this test don't work with any expected variant while if we omit it we are loosing this info and can't distinguish 2 params data set without expectation and 1 param with expectation. And all the info we that remains is stored in the name of last argument in lambla, which can be named any way developer wants right now -> this can lead to tests that will be really hard to read.

artem-zinnatullin:
> don't really see the value of having an extra expected parameter. Just by looking at the lambda parameters, I can clearly tell what each parameters mean.

Well, this is true if you already have that code, but in order to write it you need to understand how to pass expected result — better to have named argument…?

>And all the info that remains is stored in the name of last argument in lambla, which can be named any way developer wants right now (which is really common is teams with several members) → knowledge is absolutely implicit → no strict convention → harder to maintain and support.

100% agree.

@raniejade what cons of having named argument do you see except for `_` last parameter? It's optional for the caller, clear for maintainers, users and in the kotlindoc. I mean, it looks like win-win 😸 :

* Users who want explicitness can get it by specifying named argument.
* Users who want  less verbosity can omit it and just pass a value.

Alternative solution — 2 separate set of functions, `data()` for data sets without expected values, `edata()` (I'm bad at naming) for data sets with expected values.

>you still have n + 1 lambda parameters.

Is something we can discuss with Kotlin team and make a feature request to fix this on compiler side so it won't require to declare unused tail-parameters.

raniejade:
@artem-zinnatullin @igor-korotenko point taken :) I only considered properly written tests (properly named parameters, etc ...), not scenarios when you have to maintain the opposite.

> Alternative solution — 2 separate set of functions, data() for data sets without expected values, edata() (I'm bad at naming) for data sets with expected values.

```kotlin
fun <T1, Expected> SpecBody.on(description: String, vararg with: Data1<T1, Expected>, body: (T1, Expected) -> Unit)
fun <T1> SpecBody.on(description: String, vararg with: Data1<T1, Unit>, body: (T1) -> Unit)
```
Though, I'm not sure if we should provide variants.

FaustXVI:
Hello and sorry for taking so long to answer.
I like the idea of using `on` !
Having input and output in the examples is still, on most cases, an anti pattern to me. Letting the possibility for it is essential to me, but enforcing it seems wrong to me. It also assume there is only one expected value and not more (intermediate check like the size of a list before checking it's content).
I'm on my phone right now but I can provide you examples of code of when I think using expected values is a bad idea (fizzbuzz, bowling,...) and when I think it's a good idea (single letter roman numerals).

raniejade:
What's our consensus over this? :)

artem-zinnatullin:
Let's invite @hhariri to hear his opinion about design of the feature?

hhariri:
For clarity I'd go with having expected

raniejade:
There we go :D

spek-data-driven-extension/src/main/kotlin/org/jetbrains/spek/data_driven/namespace.kt
```diff
@@ -0,0 +1,41 @@
...
+        val (input1, input2, input3, expected) = it
+        defaultOn(description = description.format(input1, input2, input3, expected)) {
+            body(this, input1, input2, input3, expected)
+        }
+    }
+}
+
+data class Data1<I1, Expected>(val input1: I1, val expected: Expected)
+data class Data2<I1, I2, Expected>(val input1: I1, val input2: I2, val expected: Expected)
+data class Data3<I1, I2, I3, Expected>(val input1: I1, val input2: I2, val input3: I3, val expected: Expected)
```
artem-zinnatullin: @igor-korotenko should we add `Data…9`?

spek-data-driven-extension/src/test/kotlin/org/jetbrains/spek/data_driven/DataDrivenSpec.kt
```diff
@@ -0,0 +1,45 @@
+package org.jetbrains.spek.data_driven
+
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.given
+import org.jetbrains.spek.api.dsl.it
+import kotlin.test.assertEquals
+
+class DataDrivenSpec : Spek({
```
artem-zinnatullin: I think we need to add tests for DataDriven thing to check that it actually executes tests, kinda like what @raniejade does in such PRs https://github.com/JetBrains/spek/pull/184

artem-zinnatullin (changes requested):
Few questions

spek-data-driven-extension/src/test/kotlin/org/jetbrains/spek/data_driven/DataDrivenSpec.kt
igor-korotenko: Hi, sorry for delay, will try to accomplish this today.

spek-data-driven-extension/src/main/kotlin/org/jetbrains/spek/data_driven/namespace.kt
igor-korotenko: Hi, sorry for delay, will try to accomplish this today.

igor-korotenko: done.

artem-zinnatullin: JESOS

artem-zinnatullin (approved):

raniejade:
As per discussion, we'll include it in the `1.1` release which will be this weekend 👍

joelhandwell:
Where is documentation for data driven testing in the [spek website](http://spekframework.org/)? Sorry I could not find.

ianbrandt:
@joelhandwell, I added #364 regarding the lack of documentation for this pull request.

PR #183, opened 2017-02-19
Ensure fixtures are always executed last

### LINKED ISSUE #178: Problem with subjects (memoized) and fixtures
One thing I noticed when using subjects and fixtures (`after` variants specifically), is that the subject is *cleared* when the fixture is invoked.

```kotlin
class SimpleSpec: SubjectSpek<World<GameState>>({
    val initialStateHandler = memoized { InitialStateHandler() }
    val movementSystem = memoized { MovementSystem() }
    subject {
        createWorld<GameState> {
            initialState(GameState.Initial)
            registerStateHandler(GameState.Initial, initialStateHandler())
            registerSystem(movementSystem())
        }
    }

    beforeGroup {
        subject.setup()
    }

    afterGroup {
        // subject instance is new
        subject.cleanup()
    }

    on("update") {
        subject.update(5.0)


        it("there should be no entities in the origin") {
            with(initialStateHandler()) {
                entities.forEach {
                    expect(positionComponentMapper.get(it)).satisfy {
                        it!!.x != 0.0f && it.y != 0.0f && it.z != 0.0f
                    }
                }
            }
        }
    }
})

```

By the time `afterGroup` is invoked the subject has been cleared, so  using `subject` will return a new instance. Depending on the subject, it can cause a variety of problems. In my case an `NPE` is thrown since `setup` was not invoked.

There are 2 solutions:
1. Since fixtures and memoized (subjects are built on top of this) register a `LifecycleListener` to do their bidding (see FixturesAdapter and LifecycleAwareAdapter), we could just ensure that fixtures are always executed last.

2. The reason I used `afterGroup` was to do some cleanup for my subject, so we could support a "cleanup" callback for memoized (subject). This will be invoked when a subject is being cleared. This can also used when implementing `resource`, `Closeable#close` will be invoked in that callback (see #101).

    ```kotlin
       val myResource by resource(...)
    ```

raniejade:
Resolves #178.

### DIFF
--- a/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
+++ b/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
@@ -4,16 +4,18 @@ import org.jetbrains.spek.api.lifecycle.ActionScope
 import org.jetbrains.spek.api.lifecycle.GroupScope
 import org.jetbrains.spek.api.lifecycle.LifecycleListener
 import org.jetbrains.spek.api.lifecycle.TestScope
-import java.util.HashSet
+import java.util.ArrayList
 
 /**
  * @author Ranie Jade Ramiso
  */
 class LifecycleManager {
-    private val listeners = HashSet<LifecycleListener>()
+    private val listeners = ArrayList<LifecycleListener>()
 
     fun addListener(listener: LifecycleListener) {
-        listeners.add(listener)
+        if (!listeners.contains(listener)) {
+            listeners.add(0, listener)
+        }
     }
 
     fun removeListener(listener: LifecycleListener) {
--- a/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
+++ b/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
@@ -13,9 +13,12 @@ class LifecycleManager {
     private val listeners = ArrayList<LifecycleListener>()
 
     fun addListener(listener: LifecycleListener) {
-        if (!listeners.contains(listener)) {
-            listeners.add(0, listener)
+        if (listeners.contains(listener)) {
+            throw IllegalArgumentException("You can only register a listener once.")
+
         }
+
+        listeners.add(0, listener)
     }
 
     fun removeListener(listener: LifecycleListener) {
--- a/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
+++ b/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
@@ -1,0 +1,153 @@
+package org.jetbrains.spek.engine
+
+import com.natpryce.hamkrest.assertion.assertThat
+import com.natpryce.hamkrest.equalTo
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.lifecycle.GroupScope
+import org.jetbrains.spek.api.lifecycle.LifecycleListener
+import org.jetbrains.spek.api.lifecycle.TestScope
+import org.jetbrains.spek.engine.support.AbstractSpekTestEngineTest
+import org.junit.jupiter.api.Test
+
+/**
+ * @author Ranie Jade Ramiso
+ */
+class LifecycleListenerOrderTest: AbstractSpekTestEngineTest() {
+
+    @Test
+    fun testBeforeExecuteTest() {
+        buffer.setLength(0)
+
+        class TestSpek: Spek({
+            registerListener(object: LifecycleListener {
+                override fun beforeExecuteTest(test: TestScope) {
+                    buffer.appendln("1")
+                }
+            })
+
+            beforeEachTest { buffer.appendln("2") }
+
+            group("some group") {
+                beforeEachTest { buffer.appendln("3") }
+
+                test("do something") { }
+                test("do another thing") { }
+            }
+        })
+
+        executeTestsForClass(TestSpek::class)
+        val expected = """
+        1
+        2
+        3
+        1
+        2
+        3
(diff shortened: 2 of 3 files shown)


spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
```diff
@@ -4,16 +4,18 @@ import org.jetbrains.spek.api.lifecycle.ActionScope
...
 /**
  * @author Ranie Jade Ramiso
  */
 class LifecycleManager {
-    private val listeners = HashSet<LifecycleListener>()
+    private val listeners = ArrayList<LifecycleListener>()
 
     fun addListener(listener: LifecycleListener) {
-        listeners.add(listener)
+        if (!listeners.contains(listener)) {
```
artem-zinnatullin: Shouldn't we remove listener and add it as first if it was already in the list?

spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.jetbrains.spek.engine.support.AbstractSpekTestEngineTest
+import org.junit.jupiter.api.Test
+
+/**
+ * @author Ranie Jade Ramiso
+ */
+class LifecycleListenerOrderTest: AbstractSpekTestEngineTest() {
+
+    @Test
+    fun testBeforeExecuteTest() {
```
artem-zinnatullin: nit: `beforeExecuteTest`, no real need for `test` prefix :)

spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
```diff
@@ -0,0 +1,153 @@
...
+        1
+        2
+        3
+        """.trimIndent()
+
+        assertThat(buffer.toString().trimIndent(), equalTo(expected))
+    }
+
+    @Test
+    fun testAfterExecuteTest() {
```
artem-zinnatullin: nit: `-test` prefix

spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
```diff
@@ -0,0 +1,153 @@
...
+            group("some group") {
+                afterEachTest { buffer.appendln("3") }
+
+                test("do something") { }
+                test("do another thing") { }
+            }
+        })
+
+        executeTestsForClass(TestSpek::class)
+        // afterXXX fixtures are executed from bottom to top
```
artem-zinnatullin: 👍

spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
artem-zinnatullin: Or even throw exception? Not sure if it's legal to add one listener more than once

raniejade: Previously it was using a `HashedSet`, so I thought of preserving the behaviour. It's invalid to add a listener more than once, will throw an exception. 👍

spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
raniejade: True, it's just a habit I adopted awhile back :).

raniejade:
I'll update it once I get home.

raniejade:
@artem-zinnatullin sorry for the delay, here you go.

spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/lifecycle/LifecycleManager.kt
```diff
@@ -13,9 +13,12 @@ class LifecycleManager {
     private val listeners = ArrayList<LifecycleListener>()
 
     fun addListener(listener: LifecycleListener) {
-        if (!listeners.contains(listener)) {
-            listeners.add(0, listener)
+        if (listeners.contains(listener)) {
+            throw IllegalArgumentException("You can only register a listener once.")
```
artem-zinnatullin: 👍

spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleManagerTest.kt
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    fun sameListenerRegistration() {
+        val listener = object: LifecycleListener { }
+        lifecycleManager.addListener(listener)
+
+        try {
+            lifecycleManager.addListener(listener)
+            Assertions.fail("it should throw IllegalArgumentException")
+        } catch (e: IllegalArgumentException) {
+            // pass
```
artem-zinnatullin: I'd also add check for error message, it's useful in case if in future we'll have several different cases when we'll need to throw `IAE`, up to u though :)

raniejade: Let's do it in the future :)

artem-zinnatullin: Sure, feel free to merge then! Or should I do it? Usually in projects I work on anyone can merge PR if it's approved & CI passed (we even do it automatically at work)

raniejade: I'll merge it tomorrow 👍

PR #184, opened 2017-02-19
Properly handle exceptions during discovery phase

### LINKED ISSUE #176: exception in action results in green test

raniejade:
Ignore failing groups during discovery phase. Passing groups will now
be properly discovered by Spek.

Resolves #176.

### DIFF
--- a/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/FailingGroupTest.kt
+++ b/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/FailingGroupTest.kt
@@ -1,0 +1,33 @@
+package org.jetbrains.spek.engine
+
+import com.natpryce.hamkrest.assertion.assertThat
+import com.natpryce.hamkrest.equalTo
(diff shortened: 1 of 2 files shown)


spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/FailingGroupTest.kt
```diff
@@ -0,0 +1,33 @@
...
+            group("some group") {
+                test("do something") { }
+                test("do another thing") { }
+            }
+        })
+
+        val recorder = executeTestsForClass(FailingSpec::class)
+
+        assertThat(recorder.containerFailureCount, equalTo(1))
+        assertThat(recorder.testSuccessfulCount, equalTo(2))
```
artem-zinnatullin: 👍

PR #185, opened 2017-02-26
Point out necessity of setting -ea VM option when using Kotlin's built-in or kotlin-test assertions

ubuntudroid:
Fiddled around with this one quite a bit before getting Kotlin's built-in and kotlin-test assertions to work. This is no particular issue with Spek, but we want to be friendly citizens of the internet and thus should inform our fellow devs about the importance of these three chars. Without them tests will always pass in the IDE when using this particular configuration.


raniejade:
Thanks! :+1:

PR #186, opened 2017-03-04
Introduce CachingMode.SCOPE

### LINKED ISSUE #165: memoized not obeying group scope
In 1.1 beta 3 memoized (and therefore subjectspek) will not obey the group scope. As such the object cache is cleared after every group, not only the group in which the item is declared (or top scope in case of subjectspek). This behaviour is rather surprising and should probably be fixed checking the group parameter against a field assigned on construction.

raniejade:
In this mode instances will be cached within the group the memoized
was created. This has the same semantics with `beforeGroup` and a local
variable declaration.

Resolves #165.

### DIFF
--- a/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/MemoizedTest.kt
+++ b/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/MemoizedTest.kt
@@ -13,7 +13,7 @@ import org.junit.jupiter.api.Test
  */
 class MemoizedTest: AbstractSpekTestEngineTest() {
     @Test
-    fun testMemoizedTestCaching() {
+    fun memoizedTestCaching() {
         class MemoizedSpec: Spek({
             val foo = memoized {
                 listOf(1)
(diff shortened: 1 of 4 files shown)


spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/MemoizedTest.kt
```diff
@@ -50,12 +50,118 @@ class MemoizedTest: AbstractSpekTestEngineTest() {
...
+
+
+            test("check") {
+                assertThat(memoized1, !sameInstance(memoized2))
+            }
+        })
+
+        val recorder = executeTestsForClass(MemoizedSpec::class)
+
+        assertThat(recorder.testFailureCount, equalTo(0))
```
artem-zinnatullin: nit: don't you want to also check that required amount of tests passed?

raniejade: I just want to check that there are no failures:

This should pass:
```kotlin
            test("check") {
                assertThat(memoized1, !sameInstance(memoized2))
            }
```

Issue #168, opened 2017-01-23
Add documentation on how to use a different version of kotlin


raniejade:

ivanpopelyshev:
This is real issue, I need it for 1.1, it doesnt work with dependency of kotlin-reflect-1.0.6

raniejade:
@ivanpopelyshev (gradle) this works for me:

```gradle
dependencies {
    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
    compile "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"

    compile "org.jetbrains.kotlin:kotlin-test:$kotlinVersion"
    compile ('org.jetbrains.spek:spek-api:1.1.0-beta3') {
        exclude group: 'org.jetbrains.kotlin'
    }

    testCompile 'com.winterbe:expekt:0.5.0'

    testRuntime ('org.jetbrains.spek:spek-junit-platform-engine:1.1.0-beta3') {
        exclude group: 'org.junit.platform'
        exclude group: 'org.jetbrains.kotlin'
    }
}
```

You could also use https://github.com/spring-gradle-plugins/dependency-management-plugin.

raniejade:
Also IDEA plugin is broken when using Kotlin `1.1`, but will release a fix sometime this week.

ivanpopelyshev:
I'm trying to recompile it and publish in my bintray

ivanpopelyshev:
Ok, here it is: "http://dl.bintray.com/hackerham/spek-1.1/"

All except javadocs. Built for 1.1.0-beta-18

ivanpopelyshev:
@raniejade IDEA plugin works for me just fine, with my version of spek. I didn't check it your way.

mradzinski:
@ivanpopelyshev does "http://dl.bintray.com/hackerham/spek-1.1/" has beta38 in it? I saw a ```beta3-b38``` dependency but seems to be empty.

ivanpopelyshev:
fixing it :)

ivanpopelyshev:
Cant configure the version. It just doesnt want to upload jar files :( How to name it?

ivanpopelyshev:
OK, i was stupid, sorry :) I re-created the repo from scratch, use it!

mradzinski:
@ivanpopelyshev You recreated it from scratch at the same location? ```beta3-b38``` is not available anymore :\

ivanpopelyshev:
[...]

Issue #105, opened 2016-08-04
Test Versions Differ


jpshelley:
https://kotlinlang.slack.com/archives/spek/p1470260654000104

I tried `1.0.25` but it complains that my kotlin version `1.0.3` differs from my test app (`1.0.2`) so I assume spek isn’t up to date with kotlin?

Will that happen every time Kotlin updates? Seems a bit silly.

artem-zinnatullin:
Just force your build system to use your version of Kotlin in tests. This
is a common problem and not specific to Spek.

jpshelley:
@artem-zinnatullin understandable, I've had it happen plenty of times with other dependencies as well, but since Spek relies on Kotlin (and from the same company) there should be some way of hopefully keeping those in sync so that wouldn't be needed. I have no idea how that would work tho 😅

artem-zinnatullin:
There is no such way, even if Spek will depend on Kotlin 1.0.3 your app may depend on Kotlin 1.0.2 and build will still fail

jpshelley:
Valid point, maybe just need to add documentation to the setup then to point that out. Also if the maintainers had an agreement to keep up to date by the week or something that would be nice.
ie: kotlin 1.0.3 is released, then in 7 days spek will also support it (obviously that might get tricky but would be nice to know)

artem-zinnatullin:
Kotlin 1.x is backward compatible so it shouldn't be a problem for Spek to
work on newer versions of 1.x

PR #190, opened 2017-03-13
added docs for overriding kotlin version

rabidaudio:
I ran into version conflicts but [this](168#issuecomment-274860418) on #168 solved my issue so I added it to the docs.


rabidaudio:
Looks like #105 is related as well

PR #197, opened 2017-04-08
Update JUnit dependency to M4 for JetBrains/spek#195

binkley:


raniejade:
Thanks! At a glance it looks good to me, but I'll take another in depth look. 👍

binkley:
I added 1 more commit after the PR you may want.  I noticed the `isTest` and `isContainer` methods could just delegate to the enum used for `getType`.

gradle.properties
```diff
@@ -1,5 +1,5 @@
-kotlinVersion=1.0.6
-junitPlatformVersion=1.0.0-M3
+kotlinVersion=1.1.1
```
raniejade: Can you revert this to `1.0.6`? I know it's very minor, but I prefer commits to do single change.

spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/Scope.kt
```diff
@@ -87,8 +90,9 @@ sealed class Scope(uniqueId: UniqueId, val pending: Pending, val source: TestSou
             getParent().get() as GroupScope
         }
 
-        override fun isTest() = true
-        override fun isContainer() = false
+        override fun getType() = TEST
+        override fun isTest() = type.isTest
+        override fun isContainer() = type.isContainer
```
raniejade: Move `isTest()` and `isContainer()` overrides to `Scope`, so you don't have to duplicate the code for each scope type.

spek-samples/build.gradle
```diff
@@ -3,7 +3,7 @@ apply from: "$rootDir/gradle/common/kotlin.gradle"
 apply plugin: 'org.junit.platform.gradle.plugin'
 
 junitPlatform {
-    platformVersion '1.0.0-M3'
+    platformVersion '1.0.0-M4'
```
artem-zinnatullin: Ok we definitely need to move this to separate `dependencies.gradle` file, too much duplication :) 

@raniejade I'll do this in a separate PR if you don't mind

gradle.properties
binkley: Good call -- I did that for my local work, should not have pushed :(

raniejade:
Looks good for me, I'll merge it tonight or by tomorrow. Thanks!

tw:
Any update on merging this?

raniejade:
Yeah, sorry got a bit busy with work. Will merge within this week.

PR #205, opened 2017-05-13
Fix itBehavesLike for subjects

raniejade:
SubjectSpek's will now properly override subjects of nested SubjectSpek's.


spek-subject-extension/src/main/kotlin/org/jetbrains/spek/subject/Shared.kt
```diff
@@ -10,27 +10,29 @@ import org.jetbrains.spek.subject.dsl.SubjectDsl
 import org.jetbrains.spek.subject.dsl.SubjectProviderDsl
 import kotlin.reflect.KProperty
 
+class IncludedSubjectSpek<T>(val adapter: LifecycleAware<T>, spec: Spec): SubjectProviderDsl<T>, Spec by spec {
```
artem-zinnatullin: private?

raniejade: :+1:

PR #214, opened 2017-05-31
Fix early memoized resetting.

### LINKED ISSUE #213: Memoized value is not the same before and after test
The code and its output is below. The thing is — `environment` in `beforeEachTest` and `afterEachTest` are not the same value.

```kotlin
import org.jetbrains.spek.api.Spek
import org.jetbrains.spek.api.dsl.it
import org.junit.platform.runner.JUnitPlatform
import org.junit.runner.RunWith

@RunWith(JUnitPlatform::class)
class CleanupSpec : Spek({

    val environment by memoized { Environment() }

    beforeEachTest {
        println("before each test $environment")
    }

    it("works") {}

    afterEachTest {
        println("after each test $environment")
    }

}) {

    class Environment

}
```
```
before each test net.juno.rd.CleanupSpec$Environment@1757cd72
after each test net.juno.rd.CleanupSpec$Environment@6aaceffd
```

artem-zinnatullin:
Fixes #213.

### DIFF
--- a/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
+++ b/spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
@@ -72,12 +72,12 @@ class LifecycleListenerOrderTest: AbstractSpekTestEngineTest() {
         executeTestsForClass(TestSpek::class)
         // afterXXX fixtures are executed from bottom to top
         val expected = """
-        1
         3
         2
         1
         3
         2
+        1
         """.trimIndent()
 
         assertThat(buffer.toString().trimIndent(), equalTo(expected))
@@ -138,10 +138,10 @@ class LifecycleListenerOrderTest: AbstractSpekTestEngineTest() {
 
         executeTestsForClass(TestSpek::class)
         val expected = """
-        1
         3
         1
         2
+        1
         """.trimIndent()
 
         assertThat(buffer.toString().trimIndent(), equalTo(expected))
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
@ming13 suggested to reverse execution of `after*()` methods and that indeed helps, however @raniejade I've found that we have two "separately" managed lifecyclelisteners:

* Some are managed through `LifecycleManager` and `after*()` methods were called in direct order which seems like a bug.
* Some are managed through `FixturesAdapter` and `after*()` methods are called in reversed order which is correct.

Why not manage everything through `LifecycleManager` @raniejade?

raniejade:
@artem-zinnatullin `FixturesAdapter` was present even before `LifecycleManager` was implemented. Instead of re-writing fixtures, I opted to use it as a testing ground for `LifecycleManager`. It's a core feature that's implemented as an internal extension and it's possible to extract it to it's own module just like subjects.

artem-zinnatullin:
Okay, I see, we need to discuss this before doing 2.x :)

spek-junit-platform-engine/src/test/kotlin/org/jetbrains/spek/engine/LifecycleListenerOrderTest.kt
```diff
@@ -72,12 +72,12 @@ class LifecycleListenerOrderTest: AbstractSpekTestEngineTest() {
         executeTestsForClass(TestSpek::class)
         // afterXXX fixtures are executed from bottom to top
         val expected = """
-        1
         3
         2
         1
         3
         2
+        1
```
artem-zinnatullin: To be clear, this is kinda breaking change (and below in the same file), we need to try estimate whether it can break some users specs or not

raniejade: This would affect users using memoized, subjects and the ones registering `LifecycleListener`s directly. Given that this change is the **correct** behaviour for subjects and memoized, the only thing left to check is the usage for `LifecycleListener`s .

artem-zinnatullin: Alright, let's merge it I guess?

raniejade: Fire away! :+1:

PR #226, opened 2017-07-21
implemented tabular iteration ("where" syntax like spock does)

nielsfalk:
I'm a huge fan of Spock and I'd love to see its features in Spek

### DIFF
--- a/spek-data-driven-extension/src/test/kotlin/org/jetbrains/spek/data_driven/WhereSpec.kt
+++ b/spek-data-driven-extension/src/test/kotlin/org/jetbrains/spek/data_driven/WhereSpec.kt
@@ -19,10 +19,10 @@ object WhereSpec : Spek({
             }
         }, where<String,Int,Boolean> {
             //@formatter:off
-            //"name"  / "expectedLen" / "expectPalindrome"
-            ! "Niels" / 5             / false
-            ! "Max"   / 3             / false
-            ! "Anna"  / 4             / true
+            //"Name"  II "expectedLen" I "expectPalindrome"
+              "Niels" II 5             I false
+              "Max"   II 3             I false
+              "Anna"  II 4             I true
             //@formatter:on
         })
     }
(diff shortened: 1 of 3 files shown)


spek-data-driven-extension/src/test/kotlin/org/jetbrains/spek/data_driven/WhereSpec.kt
```diff
@@ -0,0 +1,32 @@
...
+            it("has a length $expectedLen") {
+                assertEquals(expectedLen, name.length)
+            }
+            it("${expectPalindrome.isIsNot()} a palindrome") {
+                assertEquals(expectPalindrome, name.isPalindrome())
+            }
+        }, where<String,Int,Boolean> {
+            //@formatter:off
+            //"name"  / "expectedLen" / "expectPalindrome"
+            ! "Niels" / 5             / false
```
raniejade: Concerned about using `/`. How would `! 10 / 5 / 5` be interpreted?

nielsfalk: yes I'd prefer | and || to separate expectations like Spock does. But Kotlin doesn't allow to overwrite these operators jet. Maybe I and II (letters not pipe) as infix function are better.

nielsfalk:
Ok my Fork now looks like this
```
object WhereSpec : Spek({
    describe("analyze names") {
        on("%s ", { name, expectedLen, expectPalindrome ->
            it("has a length $expectedLen") {
                assertEquals(expectedLen, name.length)
            }
            it("${expectPalindrome.isIsNot()} a palindrome") {
                assertEquals(expectPalindrome, name.isPalindrome())
            }
        }, where<String,Int,Boolean> {
            //@formatter:off
            "Name"  II "expectedLen" I "expectPalindrome"
            "Niels" II 5             I false
            "Max"   II 3             I false
            "Anna"  II 4             I true
            //@formatter:on
        })
    }
})
```
shall I make a new pull request?

PR #229, opened 2017-08-01
Fix outdated comment for 'on' and 'xon'.

yunikkk:


yunikkk:
@artem-zinnatullin any comments/obstacles here?..

artem-zinnatullin:
Nope :)

PR #244, opened 2017-08-06
Rename packages

### LINKED ISSUE #242: Rename packages
Related to #241, we need to rename our current package prefix `org.jetbrains.spek` to `org.spekframework`.

raniejade:
Resolves #242.

At this point IDEA support is broken since the plugin uses `org.jetbrains.spek`.


raniejade:
Merged manually to `2.x`.

PR #249, opened 2017-08-17
Add meta annotations for IDE consumption

### LINKED ISSUE #246: Introduce meta annotations for IDE consumption
Motivation
========

The IDEA plugin hard codes all synonyms we currently provide, which is not flexible if we want to support running user implemented synonyms. 

Proposal
=======
Introduce meta annotations that any IDE can check for synonyms.

```kotlin
enum class SynonymType {
    Group,
    Action,
    Test
}

annotation class Synonym(val type: SynonymType, 
                         val prefix: String = "", 
                         val excluded: Boolean = false)
```

So with the annotations above, let's annotate `describe`:

```kotlin
@Synonym(type = SynonymType.Group, prefix = "describe ")
fun SpecBody.describe(description: String, body: SpecBody.() -> Unit) {
    group("describe $description", body = body)
}

```

and `xdescribe`:

```kotlin
@Synonym(type = SynonymType.Group, prefix = "describe ", excluded = true)
fun SpecBody.xdescribe(description: String, reason: String? = null, body: SpecBody.() -> Unit) {
    group("describe $description", Pending.Yes(reason), body = body)
}
```

raniejade:
Resolves #246

### DIFF
--- a/spek-dsl/src/main/kotlin/org/spekframework/spek2/meta/Synonyms.kt
+++ b/spek-dsl/src/main/kotlin/org/spekframework/spek2/meta/Synonyms.kt
@@ -1,0 +1,12 @@
+package org.spekframework.spek2.meta
+
+enum class SynonymType {
+    Group,
+    Action,
+    Test
+}
+
+@Target(AnnotationTarget.FUNCTION)
+annotation class Synonym(val type: SynonymType,
+                         val prefix: String = "",
+                         val excluded: Boolean = false)
--- a/spek-dsl/src/main/kotlin/org/spekframework/spek2/meta/Synonyms.kt
+++ b/spek-dsl/src/main/kotlin/org/spekframework/spek2/meta/Synonyms.kt
@@ -14,7 +14,7 @@ enum class SynonymType {
  *
  *
  * @property type type of scope.
- * @property prefix prefix appended to the description, if applicable.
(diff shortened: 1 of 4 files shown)


spek-dsl/src/main/kotlin/org/spekframework/spek2/meta/Synonyms.kt
```diff
@@ -0,0 +1,12 @@
+package org.spekframework.spek2.meta
+
+enum class SynonymType {
+    Group,
+    Action,
+    Test
+}
+
+@Target(AnnotationTarget.FUNCTION)
+annotation class Synonym(val type: SynonymType,
```
artem-zinnatullin: Please add kotlindoc :) (for parameters as well)

spek-dsl/src/main/kotlin/org/spekframework/spek2/meta/Synonyms.kt
```diff
@@ -1,11 +1,22 @@
...
     Action,
     Test
 }
 
+/**
+ * Marks a function as a synonym to either a group, action or test scope.
+ *
+ *
+ * @property type type of scope.
+ * @property prefix prefix appended to the description, if applicable.
```
artem-zinnatullin: ", appears in report."?

raniejade:
@artem-zinnatullin following this up :)

PR #259, opened 2017-09-10
Rework path creation and serialization

raniejade:
Simply path serialization, encode `name` using base64 instead of escaping. Deserialization is now much simpler, just have to use `String.split` and decode each path parts.

### DIFF
--- a/spek-runtimes/common/src/main/kotlin/org/spekframework/spek2/runtime/scope/Path.kt
+++ b/spek-runtimes/common/src/main/kotlin/org/spekframework/spek2/runtime/scope/Path.kt
@@ -18,3 +18,9 @@ val Path.isRoot: Boolean
 // 1: my.package/MyClass -> my.package/MyClass/description
 // 2: my.package/MyClass/description -> my.package/MyClass
 fun Path.isRelated(path: Path) = this.isParentOf(path) || path.isParentOf(this)
+
+
+interface PathBuilder {
+    fun append(name: String): PathBuilder
+    fun build(): Path
+}
--- a/spek-runtimes/jvm/src/main/kotlin/org/spekframework/spek2/jvm/JvmPath.kt
+++ b/spek-runtimes/jvm/src/main/kotlin/org/spekframework/spek2/jvm/JvmPath.kt
@@ -2,14 +2,20 @@ package org.spekframework.spek2.jvm
 
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.runtime.scope.Path
+import org.spekframework.spek2.runtime.scope.PathBuilder
+import java.util.Base64
 import kotlin.reflect.KClass
 
-data class JvmPath private constructor(override val name: String, override val parent: JvmPath?): Path {
+private data class JvmPath constructor(override val name: String, override val parent: JvmPath?): Path {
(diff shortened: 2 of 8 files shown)


spek-runtimes/jvm/src/main/kotlin/org/spekframework/spek2/jvm/JvmPath.kt
```diff
@@ -29,39 +35,60 @@ data class JvmPath private constructor(override val name: String, override val p
...
+                // this will be an empty string
                 path.name
             } else {
-                "${serialize(path.parent)}${PATH_SEPARATOR}${path.name}".trimStart(*PATH_SEPARATOR.toCharArray())
+                "${serialize(path.parent)}$PATH_SEPARATOR${path.encoded}".trimStart(*PATH_SEPARATOR.toCharArray())
             }
         }
+
+        fun encode(name: String): String {
+            return Base64.getEncoder()
```
artem-zinnatullin: 👍

spek-runtimes/common/src/main/kotlin/org/spekframework/spek2/runtime/scope/Path.kt
```diff
@@ -18,3 +18,9 @@ val Path.isRoot: Boolean
 // 1: my.package/MyClass -> my.package/MyClass/description
 // 2: my.package/MyClass/description -> my.package/MyClass
 fun Path.isRelated(path: Path) = this.isParentOf(path) || path.isParentOf(this)
+
+
+interface PathBuilder {
+    fun append(name: String): PathBuilder
```
artem-zinnatullin: Maybe we should have `fun package()`, `fun class()`, etc instead of super generic `append()`? I'm just afraid such generic `String` based api will be easy to misuse :)

raniejade: Wouldn't that just complicate things? I would have to add intermediate interfaces:

```kotlin
interface ClassPathBuilder {
    fun scope(scope: String): ClassPathBuilder
    fun builder(): Path
}

interface PackagePathBuilder {
    fun class(cls: String): ClassPathBuilder
}

interface PathBuilder {
    fun package(pkg: String):  PackagePathBuilder
}
```
(ignore the naming for now)

artem-zinnatullin: Well, okay, but some kind of validation of the resulting string will be great then

Anyway, this is internal API so we can change it if it'll cause some problems :)

raniejade: What do you mean by validation?

artem-zinnatullin: That path points to some valid thing like class, or method because atm it's just a string builder :)

raniejade: It won't contain just a package and a class, it will also contain the scope descriptions. `com.example/SomeSpek/describe something/it should do this`.

PR #260, opened 2017-09-10
Update JUnit Platform 1.0.0-RC2 → 1.0.0, JUnit Jupiter 5.0.0-RC2 → 5.0.0.

artem-zinnatullin:
Release Notes http://junit.org/junit5/docs/current/user-guide/#release-notes-5.0.0

### DIFF
--- a/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/SpekTestEngine.kt
+++ b/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/SpekTestEngine.kt
@@ -41,7 +41,7 @@ class SpekTestEngine: HierarchicalTestEngine<SpekExecutionContext>() {
     val defaultInstanceFactory = object: InstanceFactory {
         override fun <T: Spek> create(spek: KClass<T>): T {
             return spek.objectInstance ?: spek.constructors.first { it.parameters.isEmpty() }
-                    .call()
+                .call()
         }
     }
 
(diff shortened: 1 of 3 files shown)


spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/SpekTestEngine.kt
```diff
@@ -41,7 +41,7 @@ class SpekTestEngine: HierarchicalTestEngine<SpekExecutionContext>() {
     val defaultInstanceFactory = object: InstanceFactory {
         override fun <T: Spek> create(spek: KClass<T>): T {
             return spek.objectInstance ?: spek.constructors.first { it.parameters.isEmpty() }
-                .call()
+                    .call()
```
raniejade: Minor stuff, you can ignore this. We should probably agree on the code style at some point.

artem-zinnatullin: I've tried :( IntelliJ just does that, I can edit it in text editor if you want

raniejade: You just need to set your trailing indent to 4 spaces in the coding style in IDEA, I think the default is 8.

artem-zinnatullin:
Formatting foxed!

PR #264, opened 2017-09-21
Update Gradle version to 4.2

udalov:
Gradle 3.1, which was previously used, is incompatible with the latest
Kotlin. Updating to the latest Kotlin (1.1.50) would result in an
exception during configuration phase


udalov:
All changes in this commit were done by the command

```
gradle wrapper --gradle-version 4.2
```

PR #293, opened 2017-10-22
Use api/implementation declarations, update deps management plugin 0.6.1.RELEASE → 1.0.3.RELEASE.

artem-zinnatullin:
- Documentation https://docs.gradle.org/4.2.1/userguide/java_library_plugin.html
- Separate `dependencies` blocks for `expectedBy`, `compile/api/implementation` and `test*`

This should make dependency resolution better and hide implementation details from consumers.


spek-runners/junit5/build.gradle
```diff
@@ -30,3 +21,13 @@ task javadocJar(type: Jar, dependsOn: dokka) {
     from dokka.outputDirectory
     classifier = 'javadoc'
 }
+
+dependencies {
+    compile project(':spek-runtimes:jvm')
+    compileOnly 'org.junit.platform:junit-platform-engine'
```
artem-zinnatullin, line 27: @raniejade this one worries me, it should be `api` or `implementation`, otherwise it'll crash with `ClassNotFoundError` if user won't manually add it (see below)

Or we expect users to always manually add `junit-platform-engine` as test dependency to their projects?

spek-runners/junit5/build.gradle
```diff
@@ -30,3 +21,13 @@ task javadocJar(type: Jar, dependsOn: dokka) {
...
+}
+
+dependencies {
+    testImplementation 'com.nhaarman:mockito-kotlin'
+    testImplementation 'org.junit.platform:junit-platform-engine'
```
artem-zinnatullin, line 32: This is the only reason tests for `:spek-runners:junit5` work

artem-zinnatullin:
No changelog for plugin, here is diff https://github.com/spring-gradle-plugins/dependency-management-plugin/compare/v0.6.1.RELEASE...v1.0.3.RELEASE

spek-runners/junit5/build.gradle
raniejade: We expect users to add them manually, didn't want to be bound to a specific version of JUnit 5 hence `compileOnly`.

raniejade:
I like the idea of using separate `dependencies` block. :+1:

raniejade (approved):

spek-runners/junit5/build.gradle
artem-zinnatullin: oook

arturdryomov (approved):

PR #295, opened 2017-10-22
Make 'dsl:common' transitive for 'dsl:jvm'.

artem-zinnatullin:
Otherwise `samples` project (and probably real users, we'll need to check `pom`) won't be able to write tests in IDE (`spek-dsl:common` is not in IDE test classpath in `samples` project)

screen shot 2017-10-22 at 00 14 22

But what's interesting is that it works from command line, so I guess `kotlin-platform-jvm` doesn't properly generate dependency tree and IDE doesn't see project declared as `expectedBy` (I'm not sure it should, but on first sight it seems that that would be expected)

### DIFF
--- a/spek-dsl/jvm/build.gradle
+++ b/spek-dsl/jvm/build.gradle
@@ -15,3 +15,7 @@ task javadocJar(type: Jar, dependsOn: dokka) {
 dependencies {
     expectedBy project(':spek-dsl:common')
 }
+
+dependencies {
+    compile project(':spek-dsl:common')
+}


raniejade:
You just need to update to the latest Kotlin 1.2 beta plugin for IntelliJ. :)

artem-zinnatullin:
Ahhh, damn version matching! We should move Spek, Kotlin and IntelliJ into single monorepo and build everything from sources to always stay in sync :trollface:

raniejade:
I'm thinking of moving the source of IJ plugin here. The only thing that is blocking me is `gradle-intellij-plugin` downloads an IDE in the configuration phase and it will slow down the build. I guess we could take advantage of gradle's configuration on-demand feature, but I haven't look into it yet much further.

artem-zinnatullin:
How slow is it? Additional 1-3 minutes is fine, especially if it's cached for local development

I'll be happy to see IntelliJ plugin in this repo, this will ease so many things :)

raniejade:
It downloads IntelliJ IDEA, depending on your connection it might be slow.

artem-zinnatullin:
Nah it's fine, let's create an issue and discuss it with @ming13 and @hhariri then!

raniejade:
Sure! Can you create it? I'm going out for a bit, otherwise I can do it later.

artem-zinnatullin:
Done, #298!

PR #271, opened 2017-09-27
Add docs for setting up android projects

raniejade:

PR #296, opened 2017-10-22
Update android doc.

artem-zinnatullin:
Reopen of #271

### DIFF
--- a/documentation/src/asciidoc/setting-up.adoc
+++ b/documentation/src/asciidoc/setting-up.adoc
@@ -163,6 +163,7 @@ JUnit Platform does not support android projects out of the box, we need to use 
 buildscript {
     ...
 
+    // 1. Add the android-junit5 Gradle plugin as a classpath dependency
     dependencies {
         classpath "de.mannodermaus.gradle.plugins:android-junit5:${junitPlatformVersion}"
     }
@@ -174,18 +175,19 @@ buildscript {
 .app/build.gradle
 ----
 ...
-
+// 2. Apply the android-junit5 plugin alongside your app's plugins
 apply plugin: "de.mannodermaus.android-junit5"
 
 android {
     ...
-
+    // 3. Extend the Java source set with the Kotlin folders
     sourceSets {
         main.java.srcDirs += 'src/main/kotlin'
         test.java.srcDirs += 'src/test/kotlin'
     }
 }
 
+// 4. Configure spek's TestEngine to be detected by JUnit 5
 junitPlatform {
     filters {
         engines {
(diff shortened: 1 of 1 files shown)


documentation/src/asciidoc/setting-up.adoc
```diff
@@ -152,3 +152,82 @@ class CalculatorSpec: Spek({
...
+
+[source,groovy,subs="attributes"]
+.build.gradle
+----
+buildscript {
+    ...
+
+    // 1. Add the android-junit5 Gradle plugin as a classpath dependency
+    dependencies {
+        classpath "de.mannodermaus.gradle.plugins:android-junit5:${junitPlatformVersion}"
```
artem-zinnatullin, line 168: Maybe explicitly add AGP 3.x?

artem-zinnatullin (commented):
LGTM

documentation/src/asciidoc/setting-up.adoc
arturdryomov: Not everybody uses it, you know. And I don’t know what you meant by that.

documentation/src/asciidoc/setting-up.adoc
```diff
@@ -152,3 +152,82 @@ class CalculatorSpec: Spek({
 IMPORTANT: As mentioned in the <<ide-support>> section, the IDEA plugin won't work if you're using the JUnit 4 runner.
 
 
+
+[[setting-up-android]]
+== Setting up Android
```
arturdryomov, line 157: As a matter of fact it did not help us using AS 3.0 RC 1.

Also, specify the AS versions supported by this.

arturdryomov (changes requested):

documentation/src/asciidoc/setting-up.adoc
mannodermaus: On a different note: The android-junit5 plugin uses an independent version from the JUnit Platform, so it shouldn't re-use the `$junitPlatformVersion` variable.

documentation/src/asciidoc/setting-up.adoc
raniejade: Is it safe to assume almost everybody is now using AS 3.0+?

arturdryomov: Not even close unfortunately (or otherwise). Especially noting that it is still a preview.

raniejade: Okay, the only thing that is specific to AS 3.0+ is point `7` - I'll include it there.

documentation/src/asciidoc/setting-up.adoc
artem-zinnatullin: @ming13 well, rest of this `build.gradle` uses things like new API configurations `testImplementation` that simply won't work on AGP 2.x, so I'd maintain 3.x in the docs

arturdryomov: @artem-zinnatullin That’s quite a mnemonic-like logic. I would make it explicit.

artem-zinnatullin: @ming13 ? So we should add AGP 3.x here or I didn't understand you?

raniejade:
Closing, no intention to merge this anymore.

PR #297, opened 2017-10-22
Clarify state of the project in readme.

artem-zinnatullin:
Just a small hint for those who are confused by what's going on with the project, we should probably update website later as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,14 @@
 
 ## A Specification Framework for the JVM
 
-
-
 Visit the [web site](https://jetbrains.github.io/spek) for more documentation and information
 
+## Current State of the Project
+
+- Spek 1.x is in support mode, only bug fixes and documentation updates will be merged into 1.x branch.
+- Spek 2.x is main development branch.
+
+---
 
 Copyright (c) 2012-2017, Spek Core Team and Contributors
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,10 +5,14 @@
 
 ## A Specification Framework for the JVM
 
+Visit the [web site](https://jetbrains.github.io/spek) for more documentation and information
 
+## Current State of the Project
 
-Visit the [web site](https://jetbrains.github.io/spek) for more documentation and information
+- Spek 1.x is in support mode, only bug fixes and documentation updates will be merged into 1.x branch.
+- Spek 2.x is main development branch.
```
raniejade: Are we bringing back `master` at some point?

raniejade (approved):

README.md
artem-zinnatullin: I would say "no" for now, it's just a branch name :) Semantic versioning is easier when branches follow major version so if we would need 3.x that won't require killing master again

raniejade: :+1:

arturdryomov: Probably it would be helpful to state `2.x` goals or link to the milestone.

arturdryomov (changes requested):

README.md
artem-zinnatullin: Added link to milestone, I didn't do it because it was something more suitable for 2.x branch or website, but ok

artem-zinnatullin:
@ming13 rePTAL :)

ColinHebert:
Are there alpha/milestone versions of 2.x available? Should that be advertised?

artem-zinnatullin:
@ColinHebert not yet, you can build it yourself from `2.x` branch though, we just don't want to ship something that can have breaking changes in every update and make people angry :)

arturdryomov (approved):

Issue #191, opened 2017-03-21
CI build status for PRs


artem-zinnatullin:
Currently PRs on GitHub do not have build status check which is very dangerous, we have [TeamCity as CI](https://teamcity.jetbrains.com/viewType.html?buildTypeId=Spek_BuildAndTests) and can configure it to report build status to GitHub.

Another option would be to use Travis or CircleCI and keep their ci setup under version control in the repo.

PR #301, opened 2017-10-23
Add Travis CI config, build script.

artem-zinnatullin:


ci/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
artem-zinnatullin, line 1: This script decouples actual CI job from Travis so we could reuse it with TC

ci/build.sh
```diff
@@ -0,0 +1,7 @@
...
+set -e
+
+# You can run it from any directory.
+PROJECT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/.."
+
+"$PROJECT_DIR"/gradlew --no-daemon --info clean build
```
arturdryomov: Replace `--info` with `--stacktrace`?

arturdryomov (approved):

ci/build.sh
artem-zinnatullin: They're not the same though, `--stacktrace` doesn't increase log level, it just prints stacktrace in case of it, but lemme add it

arturdryomov: BTW, check the full log, seems like `./gradlew assemble` is executed for some reason before `build.sh`.

arturdryomov: But do we need increased log level?

artem-zinnatullin: But do we need `--stacktrace`? It will print stacktrace even if test fails for example :)

It's typical problem — you don't need too much logs unless something went wrong

raniejade (approved):
Should we also add the travis badge on README?

artem-zinnatullin:
>Should we also add the travis badge on README?

I think we're ok without it, Hadi confirmed that there is some problem with TC, maybe we'll switch back to it again later

artem-zinnatullin:
I'm merging this for now to unblock other PRs, made Travis status check required for 2.x instead of TC, this is also related to #191

PR #305, opened 2017-10-30
Override default Travis install step to avoid unnecessary './gradlew assemble'.

artem-zinnatullin:
See https://github.com/travis-ci/travis-ci/issues/8667 for details, this will speedup all CI builds!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: java
 jdk: oraclejdk8
 
 install:
-  - echo "Override default Travis install step to avoid unnecessary './gradlew assemble'."
+  - echo "Overriding default Travis install step to avoid unnecessary './gradlew assemble'."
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,6 +4,9 @@ sudo: false
 language: java
 jdk: oraclejdk8
 
+install:
+  - echo "Override default Travis install step to avoid unnecessary './gradlew assemble'."
```
arturdryomov: Overriding?

arturdryomov (approved):

.travis.yml
artem-zinnatullin: Yeah, that would read better in logs, rather than ask user to take some action

raniejade (approved):

PR #306, opened 2017-10-30
Declare DSL extensions as direct functions to simplify usage.

### LINKED ISSUE #284: Replace DSL Extensions with Public Methods
It is kind of annoying to import DSL operators semi-manually. Yes, the IDE most of the time will suggest the right thing, but it is still confusing to see DSL operators as “optional” extensions. Technically it is nice to see how it is done via extensions, the real world usability is not so great. If we all agree about DSL stability across releases this shouldn’t be an issue.

Pros.
* Less semi-manual imports which take seconds every time you write new test and is very annoying.
* Better adoption rate because of less confusion “what to import”.

Cons.
* None?

artem-zinnatullin:
Closes #284.


spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/SpecBody.kt
```diff
@@ -25,4 +25,92 @@ interface SpecBody: TestContainer {
...
+
+    /**
+     * Creates a [group][SpecBody.group].
+     *
+     * @author Ranie Jade Ramiso
+     * @since 1.0
+     */
+    @Synonym(type = SynonymType.Group, prefix = "describe")
+    fun describe(description: String, body: SpecBody.() -> Unit) {
+        group("describe $description", body = body)
```
arturdryomov, line 37: It’s a bummer that we cannot reuse `prefix` from the annotation here :-(

spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/SpecBody.kt
```diff
@@ -25,4 +25,92 @@ interface SpecBody: TestContainer {
...
+    }
+
+    /**
+     * Creates a [group][SpecBody.group].
+     *
+     * @author Ranie Jade Ramiso
+     * @since 1.0
```
arturdryomov, line 44: Maybe remove these?

arturdryomov (approved):

spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/SpecBody.kt
artem-zinnatullin: I would remove that, but it's @raniejade's child, so only if he agrees :)

spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/SpecBody.kt
artem-zinnatullin: Not a bummer, but yeah ¯\_(ツ)_/¯

raniejade (approved):
nit: Should we remove the author tags? They become unmaintained as project the goes on. (it was an old habit of mine)

spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/SpecBody.kt
raniejade: :+1: actually I just commented about it :).

artem-zinnatullin:
@raniejade #307, will address in a separate PR

PR #309, opened 2017-10-30
Delete unreachable code from FailingGroupTest.

artem-zinnatullin:
Also noticed during build:

```console
FailingGroupTest.kt: (19, 17): Unreachable code
```

Initially I suppressed it, but then decided to remove since Spek has no way to catch exception and continue method invocation there, so it actually doesn't matter if we have this code or not.

### DIFF
--- a/spek-runtimes/jvm/src/test/kotlin/org/spekframework/spek2/jvm/FailingGroupTest.kt
+++ b/spek-runtimes/jvm/src/test/kotlin/org/spekframework/spek2/jvm/FailingGroupTest.kt
@@ -16,7 +16,6 @@ class FailingGroupTest: AbstractSpekRuntimeTest() {
 
             group("some failing group") {
                 throw RuntimeException()
-                test("this won't be executed") { }
             }
 
             group("some group") {


raniejade (commented):
I wouldn't delete that as it's part of the test case. Say the runtime ignored the thrown exception, the following will still pass.

```kotlin
assertThat(recorder.containerFailureCount, equalTo(1))
assertThat(recorder.testSuccessfulCount, equalTo(2))
```

Is it possible to ignore it?

artem-zinnatullin:
@raniejade totally possible to suppress, but as I mentioned in description, it's a `throw` and Spek has no way to continue method/function/lambda/whatever execution there :)

raniejade:
Let's keep it and just suppress the error, it adds context to the test. If we remove it `assertThat(recorder.testSuccessfulCount, equalTo(2))` doesn't make sense.

artem-zinnatullin:
I'll give you time to open the code in IDE and look again, maybe you'll change your opinion :) 

Because code is indeed unreachable 😸 

But if not — I'm fine with any of the proposed solutions and will update PR tomorrow, gotta go sleep

arturdryomov (approved):

raniejade:
If you remove `assertThat(recorder.testSuccessfulCount, equalTo(2))` too, then I'm fine with it :)

raniejade:
Closing this old PR, it is way behind `2.x`.

PR #312, opened 2017-10-30
Update Gradle to version 4.3

eriwen:
This simply updates Spek to use the newly released Gradle 4.3.

PR #313, opened 2017-10-30
Adjust timestamp in JAR manifests to allow Gradle incremental builds

eriwen:
Adding a timestamp to all JAR manifests causes Gradle's incremental build to always consider all `Jar` tasks out-of-date.

What this change does is sets the `Build-Time` to the current time only if it's on TeamCity, and otherwise uses a constant value.
See https://guides.gradle.org/performance/#incremental_build and https://youtu.be/MofXNalZU-E?t=27m55s for more information.


artem-zinnatullin:
Can you please retarget/reopen this against `2.x` branch? 

See https://github.com/spekframework/spek/pull/312#issuecomment-340615556 for details :)

eriwen:
Sure, will do :)

PR #315, opened 2017-10-30
Adjust timestamp in JAR manifests to allow Gradle incremental builds

eriwen:
Adding a timestamp to all JAR manifests causes Gradle's incremental build to always consider all Jar tasks out-of-date.

What this change does is sets the Build-Time to the current time only if it's on TeamCity, and otherwise uses a constant value.
See https://guides.gradle.org/performance/#incremental_build and https://youtu.be/MofXNalZU-E?t=27m55s for more information.

### DIFF
--- a/gradle/common/kotlin-jvm.gradle
+++ b/gradle/common/kotlin-jvm.gradle
@@ -13,6 +13,6 @@ tasks.withType(Jar) { task ->
             'Implementation-Title': 'Spek', \
             'Implementation-Version': rootProject.version, \
             'Implementation-Kotlin-Version': kotlinVersion, \
-            'Build-Time': new Date().format('yyyy-MM-dd\'T\'HH:mm:ssZ')
+            'Build-Time': System.getenv("TEAMCITY_VERSION") ? new Date().format('yyyy-MM-dd\'T\'HH:mm:ssZ') : new Date(0).format('yyyy-MM-dd\'T\'HH:mm:ssZ')
     }
 }


gradle/common/kotlin-jvm.gradle
```diff
@@ -13,6 +13,6 @@ tasks.withType(Jar) { task ->
             'Implementation-Title': 'Spek', \
             'Implementation-Version': rootProject.version, \
             'Implementation-Kotlin-Version': kotlinVersion, \
-            'Build-Time': new Date().format('yyyy-MM-dd\'T\'HH:mm:ssZ')
+            'Build-Time': System.getenv("TEAMCITY_VERSION") ? new Date().format('yyyy-MM-dd\'T\'HH:mm:ssZ') : new Date(0).format('yyyy-MM-dd\'T\'HH:mm:ssZ')
```
artem-zinnatullin, line 16: I will most likely edit it to not directly depend on TeamCity and maybe produce build time with at least "current day" resolution for local builds, 

but I don't wanna block you :)

artem-zinnatullin (approved):

artem-zinnatullin:
We're not really suffering from this jar repackaging problem, especially since it shouldn't invalidate code compilation, but thanks!

Btw, for Android projects you might want to check out https://github.com/zsoltk/paperwork Gradle Plugin (see its README, many Android projects suffer from similar problem of putting time/gitSha into `BuildConfig` class at build time)

gradle/common/kotlin-jvm.gradle
eriwen: You won't block anything :) 

Just want this to work for everyone first and foremost. Any boolean that indicates that it's a local build (vs CI) will work here. 

BTW, I just ran into this while trying out Kotlin support in Gradle and figured to make a quick fix.

eriwen:
Your suggestion for paperwork is a wonderful one! We want something like this for everyone — thank you!

PR #316, opened 2017-10-31
Re-enable distribution

raniejade:
Cleaned up distribution scripts and the generated pom files. To checkout it out run: `./gradlew publishToMavenLocal`. Also renamed the following directories to be consistent:

- `spek-runtimes` -> `spek-runtime`
- `spek-extensions` -> `spek-extension`
- `spek-runners` -> `spek-runner`


raniejade:
Looks like there is an issue with travis: https://www.traviscistatus.com/incidents/xgpn2tq6mx7k

distribution/build.gradle
```diff
@@ -1,280 +1,86 @@
...
 
-            pom.withXml {
-                asNode().dependencies.dependency.each {
-                    it.scope[0].value = ('compile')
-                }
+        pom.withXml {
+            asNode().children().last() + {
+                resolveStrategy =  DELEGATE_FIRST
+                name 'Spek'
+                description 'A Specification Framework for the JVM'
```
artem-zinnatullin: Let's update this? It's more for "Kotlin" now :)

artem-zinnatullin (approved):

PR #339, opened 2017-12-02
list Atrium, an assertion library

robstoll:
Would be nice if you could list Atrium in the list of assertion libraries. Atrium is using Spek itself and I regularly report bugs


raniejade:
Can you do it against `1.x` in this file https://github.com/spekframework/spek/blob/1.x/documentation/src/asciidoc/writing-specifications.adoc? `gh-pages` branch is generated from  `1.x`.

robstoll:
sure, I'll create a new pull request => #340

PR #342, opened 2017-12-11
Fix cross-module groupId/artifactIds in pom.

### LINKED ISSUE #317: Invalid dependency declaration in pom
I missed this out previously, looking a way to fix it without changing the project's `group`. 

```
    <dependency>
      <groupId>spek.spek-dsl</groupId>
      <artifactId>common</artifactId>
      <version>1.1.5</version>
      <scope>compile</scope>
    </dependency>
```

artem-zinnatullin:
This PR fixes #317.

Previously Gradle screwed up our cross-module dependencies in generated pom:

```xml
<dependency>
  <groupId>spek.spek-runtime</groupId>
  <artifactId>jvm</artifactId>
  <version>1.1.5</version>
  <scope>compile</scope>
</dependency>
```

After fix in this PR:

```xml
<dependency>
  <groupId>org.spekframework.spek2</groupId>
  <artifactId>spek-dsl-jvm</artifactId>
  <version>1.1.5</version>
  <scope>compile</scope>
</dependency>
```

I've tried different ways to fix it, but nothing worked as planned so I ended up just rewriting pom xml.

Thanks Gradle

me

### DIFF
--- a/distribution/build.gradle
+++ b/distribution/build.gradle
@@ -7,16 +7,17 @@ apply from: "$rootDir/gradle/common/dependencies.gradle"
 description = 'Spek Distribution packages'
 
 
-def artifact(publication, project, artifactName) {
+def artifact(Publication publication, Project project, artifactName) {
     def configure = {
         groupId = 'org.spekframework.spek2'
         artifactId = artifactName
         version = rootProject.version
 
         from project.components.java
+        
         artifact project.tasks['sourceJar']
         artifact project.tasks['javadocJar']
-
+        
         pom.withXml {
             asNode().children().last() + {
                 resolveStrategy =  DELEGATE_FIRST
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
What I've tried:

- `includeFlat` instead of `include` (not sure what I expected from it, [documentation is pretty vague](https://docs.gradle.org/current/userguide/build_lifecycle.html#sec:flat_layouts)) 
- Override `groupId` on early configuration stages in subprojects
- Hook into `artifactId` resolution before pom generation

Would be great if we could define custom artifactId on any java or java-library projects, but we can only define `version` and `groupId`.

raniejade (approved):
LGTM :+1:

distribution/build.gradle
```diff
@@ -7,16 +7,17 @@ apply from: "$rootDir/gradle/common/dependencies.gradle"
...
         groupId = 'org.spekframework.spek2'
         artifactId = artifactName
         version = rootProject.version
 
         from project.components.java
+        
         artifact project.tasks['sourceJar']
         artifact project.tasks['javadocJar']
-
+
```
vanniktech: nit: remove trailing whitespace

vanniktech (approved):

jaredsburrows:
This can work but it will be hard to maintain. See my comments here: https://github.com/spekframework/spek/issues/317#issuecomment-350780453.

distribution/build.gradle
artem-zinnatullin: ahh thx

artem-zinnatullin:
It works and it's not hard to maintain :)

PR #350, opened 2018-01-10
Update to Gradle 4.4.1

raniejade:


raniejade:
@artem-zinnatullin

artem-zinnatullin (approved):

artem-zinnatullin:
Interesting, looks like Dokka's poms are breaking plugin resolution since Gradle 4.4 https://github.com/Kotlin/dokka/issues/146#issuecomment-350735405

raniejade:
A workaround is also mentioned in the issue, I'll see if it works.

artem-zinnatullin:
Looks like there is no good solution until Dokka will fix it on their side and release new version

artem-zinnatullin:
There is also a comment stating that workaround doesn't work though

raniejade:
Oh well, we will just have to wait.

lptr:
I submitted #356 that works around the Dokka issue and also upgrades to Gradle 4.4.1.

artem-zinnatullin:
Closing in favor of #356

PR #351, opened 2018-01-10
Publish 2.x snapshots

raniejade:
Use artifactory to publish to `oss.jfrog.org`, also removed `git-versioning` and manually set version to `2.0.0-SNAPSHOT`.


raniejade:
@artem-zinnatullin

raniejade:
@artem-zinnatullin wanna take a look and take over?

artem-zinnatullin (approved):

artem-zinnatullin:
Looks good to me, but we don't invoke publishing tasks on CI, I can do this in a separate PR if it sounds good to you

raniejade:
Sure, go ahead. I can provide `bintrayUser` and `bintrayApiKey` properties, these are my personal keys as bintray does not have organizational api keys.

artem-zinnatullin:
@raniejade sounds good, but let's do it through Travis encrypted env variables? I'll prepare a PR and you'll just need to follow instructions on how to set them on Travis

PR #358, opened 2018-01-21
[Don't merge yet] Publish snapshots on 2.x branch CI builds.

artem-zinnatullin:
This will publish snapshots on 2.x branch builds on CI (I need to update branch name before merge)

@raniejade can you please go into Travis CI settings for Spek project https://travis-ci.org/spekframework/spek/settings and add two environment variables:

- `BINTRAY_USER`
- `BINTRAY_API_KEY`

Once you add them noone will be able to see them in Travis settings, please make sure you don't check "Display value in build log".

Important note: we will need to carefully review changes to build scripts (Bash and Gradle) because someone can try to change them to print/leak those variables.

@vanniktech would be great if you could take a look as well


raniejade (approved):
LGTM 👍 

I've already added the env variables in travis.

ci/build.sh
```diff
@@ -5,3 +5,28 @@ set -e
...
+    fi
+}
+
+# TODO don't merge, change branch name to '2.x' before merge.
+if [ "$GIT_BRANCH" == "az/snapshot-pub-creds" ]; then
+    checkIfEnvVarIsNotEmpty "BINTRAY_USER"
+    checkIfEnvVarIsNotEmpty "BINTRAY_API_KEY"
+
+    echo "Publishing snapshot..."
+    "$PROJECT_DIR"/gradlew artifactoryPublish --info --stacktrace
```
vanniktech: Do we want `--no-daemon` here too?

vanniktech (approved):
Looks good.

I usually put the credentials encrypted into the configuration file - https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

ci/build.sh
artem-zinnatullin: So, I was thinking about it more and more recently and concluded that Gradle Daemon is essential part of development workflow, so maybe it's time to use it on CI (we do at work because CI pipeline is complex and does multiple calls to Gradle)

ci/build.sh
```diff
@@ -5,3 +5,30 @@ set -e
 PROJECT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/.."
 
 "$PROJECT_DIR"/gradlew --no-daemon --info --stacktrace clean build
+
+BINTRAY_USER="$BINTRAY_USER"
+BINTRAY_API_KEY="$BINTRAY_API_KEY"
+
+GIT_BRANCH=`git rev-parse --abbrev-ref HEAD`
```
artem-zinnatullin: uh, Travis checks out commit directly so we can't ask git for current branch

I'll workaround that in a way that it'll work both locally and on Travis

artem-zinnatullin:
lool, `travis-ci/pr`build sets branch name as` 2.x` wow

artem-zinnatullin:
@raniejade Artifactory returns `Unauthorized Status code: 401`, can you please checkout this branch and debug locally to check credentials?

You can run it like that 

```console
BINTRAY_USER="user" BINTRAY_API_KEY="key" ci/build.sh
```

ci/build.sh
vanniktech: Then we should also get rid of `--no-daemon` in line 9, should not we?

Zkffkah:
Sorry to interrupt here, I was wondering is there any progress on this snapshot release? It seems the authorized problem is solved, but still no snapshot version release. Any problem here? I really want to use spek2 as soon as possible. Thanks.

arturdryomov:
@lowwor, sorry for the delay, the development team is kind of preoccupied with personal thingies currently. @artem-zinnatullin will finish this, but using the Valve Time.

ci/build.sh
```diff
@@ -4,4 +4,33 @@ set -e
...
+
+BINTRAY_USER="$BINTRAY_USER"
+BINTRAY_API_KEY="$BINTRAY_API_KEY"
+
+GIT_BRANCH=`getCurrentGitBranch`
+
+function checkIfEnvVarIsNotEmpty() {
+    local env_var_name="$1"
+
+    if [ -z "env_var_name" ]; then
```
raniejade: Not a bash expert, but shouldn't this be `"$env_var_name"`?

ci/build.sh
```diff
@@ -4,4 +4,33 @@ set -e
...
+        echo "Error: environment variable $env_var_name is empty."
+        exit 1
+    fi
+}
+
+GIT_BRANCH_FOR_SNAPSHOTS="az/snapshot-pub-creds"
+
+# TODO don't merge, change branch name to '2.x' before merge.
+if [ "$GIT_BRANCH" == "$GIT_BRANCH_FOR_SNAPSHOTS" ]; then
+    checkIfEnvVarIsNotEmpty "BINTRAY_USER"
```
raniejade: `"$BINTRAY_USER"`?

ci/build.sh
```diff
@@ -4,4 +4,33 @@ set -e
...
+    checkIfEnvVarIsNotEmpty "BINTRAY_API_KEY"
```
raniejade: same as above, missing `$`?

artem-zinnatullin: facepalm

ci/build.sh
artem-zinnatullin: facepalm

artem-zinnatullin:
Still getting

```console
java.io.IOException: Failed to deploy file. Status code: 401 Response message: Artifactory returned the following errors: 
  Unauthorized Status code: 401
```

If credentials on Travis are correct, I guess the problem could be in push to snapshots repo instead of release

osipxd:
Any progress on this?

raniejade:
@artem-zinnatullin was able to fix it by changing `getProperyOrEnvVar`. I'm not sure why the previous implementation doesn't work, but it does something dodgy (printing `[][0][][0]`). Fucking, groovy/gradle magic.

```groovy
// Can be null.
String getProperyOrEnvVar(String name) {
    if (hasProperty(name)) {
        return findProperty(name)
    }
    return System.getenv(name)
}
```

raniejade:
nvm, it's still there. 😕

raniejade:
Ugghhh, there... So you can't do a method call inside `publish { ... }`. 😞

raniejade:
@artem-zinnatullin working now, please check :)

artem-zinnatullin:
ahh, I guess it's because of `project.property` call

artem-zinnatullin:
Merging, thanks @raniejade 😽

My apologies for abandoning this pr

artem-zinnatullin:
I guess I should have changed the message in resulting commit, well

PR #362, opened 2018-01-29
Fix rendering of generics in docs

### LINKED ISSUE #361: KotlinScript gradle documentation doesn't render correctly
This is what the `setting-up.adoc` has

```kotlin
configure<JUnitPlatformExtension> {
    filters {
        engines {
            include("spek")
        }
    }
}
```

However, when it gets rendered, it looks like this

```kotlin
configure {
    filters {
        engines {
            include("spek")
        }
    }
}
```

This seems to stem from the use of `subs="attributes"` in the source block. It's quite possibly an asciidoctor bug. The `<JUnitPlatformExtension>` gets rendered in HTML without escaping.

rahulsom:
When using `subs="attributes"`, asciidoctor doesn't do what is needed. It needs to be either `subs="+attributes"` or `subs="attributes+"`.

This does the former.

Fixes #361


raniejade (approved):
LGTM

PR #363, opened 2018-02-02
Data-driven extension fix

Prototik:
* Currently it just broken due to error in jvm-specific part of String.format extension (args passed as array of array instead of decoupling)
* Data* classes missing `out` variance, which makes it usage a little harded
* `data` and `on` methods have meaningless `reified` type-parameters, it's just no reason to have it
* `on` methods missing for Data4-9 arrays
* `on` methods could be slightly shorter due to destructuring Data classes in lambda declaration directly


arturdryomov:
Looks nice to me. @artem-zinnatullin, WDYT?

raniejade (approved):
LGTM

raniejade:
Merged, thanks @Prototik!

PR #367, opened 2018-02-10
Run destructor callback when memoized value leaves scope

tadfisher:
Add a `destructor` callback to handle cleanup of mutable or external state when
a `memoized()` value leaves scope. This would be useful for shared test fixtures
which require some teardown action, and extensions could allow direct adaptation
of JUnit-style "before" and "after" actions.


raniejade (approved):
LGTM, @artem-zinnatullin can you please check?.

arturdryomov:
Awesome! Double respect points for a test coverage ;-) 

Isn’t it possible to achieve a similar cleanup behavior doing `beforeEachTest` or `afterEachTest` though?

raniejade:
@ming13 It is possible but this will prevent you from mixing incompatible `CachingMode` and fixtures (e.g. `CachingMode.Group` with a `(before/after)EachTest`).

raniejade:
Merged, thanks @tadfisher!

PR #370, opened 2018-02-25
Rework dynamic test integration

raniejade:
Dynamic test is a JUnit 5 concept, to make integration easier `dynamicTestRegistered` execution event was added to `ExecutionListener`. This PR will remove that method further decoupling Spek w/ JUnit 5.

### DIFF
--- a/spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/RuntimeExecutionListenerAdapter.kt
+++ b/spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/RuntimeExecutionListenerAdapter.kt
@@ -2,16 +2,15 @@ package org.spekframework.spek2.junit
 
 import org.junit.platform.engine.EngineExecutionListener
 import org.junit.platform.engine.TestExecutionResult
-import org.spekframework.spek2.runtime.execution.ExecutionContext
+import org.spekframework.spek2.runtime.execution.ExecutionListener
 import org.spekframework.spek2.runtime.execution.ExecutionResult
-import org.spekframework.spek2.runtime.execution.RuntimeExecutionListener
 import org.spekframework.spek2.runtime.scope.ActionScopeImpl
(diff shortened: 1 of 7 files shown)


spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/RuntimeExecutionListenerAdapter.kt
```diff
@@ -23,7 +22,12 @@ class RuntimeExecutionListenerAdapter(val listener: EngineExecutionListener,
     override fun executionFinish() { }
 
     override fun testExecutionStart(test: TestScopeImpl) {
-        listener.executionStarted(test.asDescriptor())
+        val descriptor = test.asDescriptor()
+        if (test.parent is ActionScopeImpl) {
+            listener.dynamicTestRegistered(descriptor)
```
arturdryomov, line 27: Does it handle this kind of situation?

```kotlin
listOf(1,  2, 3).forEach { number ->

    context("number is [$number]") {
    
        beforeEachTest {
            println(number)
        }
    }
}
```
I’m asking mostly because of `ActionScopeImpl`.

raniejade: That should work, but I'm not sure how it relates to the PR?

arturdryomov: I’m just not seeing the `ActionScopeImpl` check in the previous version of the code.

raniejade: Ahh, your example should still work. Only affects `test`s whose parent is an `action`, which `context` is not.

arturdryomov (approved):
Merge it and let’s see how it goes!

PR #372, opened 2018-02-28
Upgrade JUnit Platform 1.1.0 / JUnit Jupiter 5.1.0

raniejade:


arturdryomov (approved):

gradle.properties
```diff
@@ -1,5 +1,5 @@
 kotlinVersion=1.2.20
-junitPlatformVersion=1.0.1
+junitPlatformVersion=1.1.0
```
arturdryomov, line 2: BTW WDYT about moving all versions to a single place?

artem-zinnatullin: PR welcome imo, not critical fro this one in particular

Create issue? :)

artem-zinnatullin (approved):
LETS MERGE

gradle.properties
raniejade: #375

Issue #328, opened 2017-11-11
Improve spec discovery


raniejade:
Currently the `jvm` runtime will search the classpath for subclasses of `Spek` during the discovery phase. This is not optimal specially when running specs from the IDE, where most of the time we only want to run a specific class. `SpekRuntime#discover` accepts a `DiscoverRequest` parameter which currently only have a single property `path`. We could add extra properties here that each runtime could use to speed up the discovery phase. One example is adding a `source directory` parameter that can limit the search space for specs.

raniejade:
cc @artem-zinnatullin @ming13 @hhariri

artem-zinnatullin:
I think we should be able to rewrite `discover()` phase of `SpekRuntime` to not start with `reflections.getSubTypesOf(Spek::class.java)` if `DiscoveryRequest` has specific path?

raniejade:
We could set a specific path for `reflections` to narrow down the search.

artem-zinnatullin:
SGTM, my point here is that initial `path` passed in `DiscoveryRequest` should be enough :)

PR #377, opened 2018-03-08
Narrow down spec discovery

raniejade:
This an attempt to solve #328. Introduced `sourceDirs` during discovery which will be used to build up `Reflections`, if empty it will re-use the current class loader's classpath.

### DIFF
--- a/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
+++ b/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
@@ -3,6 +3,6 @@ package org.spekframework.spek2.runtime.execution
 import org.spekframework.spek2.runtime.scope.Path
 import org.spekframework.spek2.runtime.scope.ScopeImpl
 
-data class DiscoveryRequest(val path: Path, val sourceDirs: List<String>)
+data class DiscoveryRequest(val path: Path, val specDirs: List<String>)
 data class DiscoveryResult(val roots: List<ScopeImpl>)
 
--- a/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
+++ b/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
@@ -3,6 +3,6 @@ package org.spekframework.spek2.runtime.execution
 import org.spekframework.spek2.runtime.scope.Path
 import org.spekframework.spek2.runtime.scope.ScopeImpl
 
-data class DiscoveryRequest(val path: Path, val specDirs: List<String>)
+data class DiscoveryRequest(val sourceDirs: List<String>, val path: Path)
 data class DiscoveryResult(val roots: List<ScopeImpl>)
 
--- a/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
+++ b/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
@@ -3,6 +3,6 @@ package org.spekframework.spek2.runtime.execution
 import org.spekframework.spek2.runtime.scope.Path
 import org.spekframework.spek2.runtime.scope.ScopeImpl
 
-data class DiscoveryRequest(val path: Path)
+data class DiscoveryRequest(val path: Path, val sourceDirs: List<String>)
 data class DiscoveryResult(val roots: List<ScopeImpl>)
 
--- a/spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
+++ b/spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
@@ -26,7 +26,7 @@ actual class SpekRuntime: AbstractRuntime() {
     }
 
     override fun discover(discoveryRequest: DiscoveryRequest): DiscoveryResult {
-        val reflections = createReflections(discoveryRequest.specDirs)
+        val reflections = createReflections(discoveryRequest.sourceDirs)
         val scopes = reflections.getSubTypesOf(Spek::class.java)
             .map(Class<out Spek>::kotlin)
             .filter { it.findAnnotation<Ignore>() == null }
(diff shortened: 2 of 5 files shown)


spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
```diff
@@ -3,6 +3,6 @@ package org.spekframework.spek2.runtime.execution
 import org.spekframework.spek2.runtime.scope.Path
 import org.spekframework.spek2.runtime.scope.ScopeImpl
 
-data class DiscoveryRequest(val path: Path)
+data class DiscoveryRequest(val path: Path, val sourceDirs: List<String>)
```
arturdryomov: Naming leads to confusion (or maybe it’s just me). Can you please give a few words about differences between `Path` and directory path as `String`?

I always thought that `Path` is a way to find out which tests to run in the current classpath. I’ve read the original issue, basically you are trying to optimize the discovery phase by narrowing down the search, which is a good thing. But isn’t it a better approach just to add more functionality to the `Path` class? I. e. `Path.getFile` or something.

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
```diff
@@ -58,4 +52,19 @@ actual class SpekRuntime: AbstractRuntime() {
             .map { it.objectInstance ?: it.primaryConstructor!!.call() }
             .firstOrNull() ?: defaultInstanceFactory
     }
+
+    private fun createReflections(testDirs: List<String>): Reflections {
+        val urls = if (testDirs.isEmpty()) {
+            ClasspathHelper.forClassLoader()
+        } else {
+            testDirs.map { java.nio.file.Paths.get(it) }
```
arturdryomov: Isn’t it better just pass `File` instances instead of `String`?

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
```diff
@@ -58,4 +52,19 @@ actual class SpekRuntime: AbstractRuntime() {
...
+                .map { it.toUri().toURL() }
+        }
+
+        return Reflections(
```
arturdryomov: BTW, why Spek does not use the native Kotlin relfection for this? Or it is too limited for Spek needs?

spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
raniejade: `Path` is used for filtering (think of XPath), it used to run a specific scope. `/com.example/FooSpec` path will only execute `com.example.FooSpec`. While `sourceDirs` is where the engine searches for specs.

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
raniejade: 👍

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
raniejade: Not limited per se, you just need to do a lot of manual tedious hacking. `Reflections` does it clear and neat way. Basically, less things to think about.

spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/execution/Discovery.kt
raniejade: I've renamed it to `specDirs` if that helps.

arturdryomov: My main concern is that `Path` already has enough metadata for narrowing down the search. It has the package name (which can be used with `Reflections.forPackages`) and the class name. Shouldn’t we use that instead of directories? Seems like you’ve introduced `Path` intentionally to avoid any file system-related associations.

BTW, what is the goal of `Path` as URI? I. e. why is it set to be `/package/class/`? As far as I can see it in the source code it is just a package + a class all the time (am I wrong?)

In any case, I think it is possible to provide extra public fields for the JVM `Path`, i. e. `package` and `class` which can be used by the JVM runtime. Similar can be done for Native and JS or whatever. Am I wrong in this?

raniejade: `Path` and `specDirs` are fundamentally different things. Firstly path is used in filtering scopes you only want to run, say you have the following spec:

```kotlin
package com.example

object MySpec: Spek({
  group("some group") {
      test("some test 1") { ... }
      test("some test 2") { ... }
  }
})
```

If you only want to execute `some test 1` you pass in a path: `/com.example/MySpec/some group/some test/`. If you want to run all specs in `com.example.` you pass in `/com.example/`. For running `MySpec` only, you pass in `/com/example/MySpec`.

`specDirs` on the other hand is the location where the engine looks for this specs. So it actually points to an actual file in the file system,  in gradle, for example, it can contain: `$buildDir/classes/kotlin/test`.

arturdryomov: Looking that way dosn’t bring much optimization then since _most of the time_ there is a single `test` directory in module. 

@artem-zinnatullin, WDYT?

raniejade: I think you are still missing the point, that's what this PR trying to do. Narrow down the discovery to `specDirs` otherwise the whole classpath will be searched for specs (that includes product and third-party classes).

arturdryomov: Ah, that makes sense now. I’m not sure it was mentioned anywhere.

Can we make `sourceDirs` a first argument then?

arturdryomov (approved):

PR #378, opened 2018-03-09
Add support for filtering multiple paths

raniejade:


spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
```diff
@@ -38,15 +38,13 @@ actual class SpekRuntime: AbstractRuntime() {
             .filter { !it.isAbstract }
             .map { klass ->
                 klass to PathBuilder.from(klass).build()
+            }.map { (klass, path) ->
```
arturdryomov: Let’s follow the same style everywhere here, i. e.

```kotlin
.map { 
}
.filter { }
.map { }
```

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
```diff
@@ -38,15 +38,13 @@ actual class SpekRuntime: AbstractRuntime() {
...
+                val matched = discoveryRequest.paths.firstOrNull { it.isRelated(path) }
+                matched to resolveSpec(instanceFactoryFor(klass).create(klass), path)
```
arturdryomov: I think that there is some redundancy here. Even if path was not matched you are calling `resolveSpec` for it.

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
raniejade: 👍

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
raniejade: 👍

raniejade:
@ming13 changes in!

arturdryomov (approved):

Issue #217, opened 2017-06-07
Memoized does not allow null values — runtime crash.


artem-zinnatullin:
```kotlin
by memoized { func() }
```

If `func()` returns `null`, Spek crashes with:

```
kotlin.KotlinNullPointerException
	at org.jetbrains.spek.engine.lifecycle.LifecycleAwareAdapter.invoke(LifecycleAwareAdapter.kt:23)
	at org.jetbrains.spek.engine.lifecycle.LifecycleAwareAdapter.getValue(LifecycleAwareAdapter.kt:17)
```

Because `null` used internally as not initialized state of memoized.

Faced this in our test suite, I'll work on fix.

PR #382, opened 2018-03-26
Memoized supports nullable values

tokou:
Fixes - #217 

Resorted to using a wrapper for the cached value.

Other explored routes :
- `lateinit var` does not work because it is not allowed to be nullable
- adding a boolean to check caching instead of comparing to null value does not work because we cannot declare cached property as `T` and initialize it to `null`

### DIFF
--- a/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/lifecycle/LifecycleAwareAdapter.kt
+++ b/spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/lifecycle/LifecycleAwareAdapter.kt
@@ -9,29 +9,39 @@ import kotlin.reflect.KProperty
 
 sealed class LifecycleAwareAdapter<T>(val factory: () -> T, val destructor: (T) -> Unit)
     : LifecycleAware<T>, LifecycleListener {
-    protected data class CachedValue<out T>(val value: T)
-    protected var cached: CachedValue<T>? = null
+
+    protected sealed class Cached {
+        object Empty : Cached()
+        data class Value<out T>(val value: T) : Cached()
+    }
+
+    protected fun <T> Cached.reified() = this as Cached.Value<T>
+
+    protected var cached: Cached = Cached.Empty
 
     override fun getValue(thisRef: Any?, property: KProperty<*>) = invoke()
 
-    override fun invoke(): T {
-        if (cached == null) {
-            cached = CachedValue(factory())
+    override fun invoke(): T = when(cached) {
+        Cached.Empty -> {
+            cached = Cached.Value(factory())
+            cached.reified<T>().value
         }
-        return cached!!.value
+        is Cached.Value<*> -> cached.reified<T>().value
     }
 
     class GroupCachingModeAdapter<T>(factory: () -> T, destructor: (T) -> Unit)
         : LifecycleAwareAdapter<T>(factory, destructor) {
-        private val stack = mutableListOf<CachedValue<T>?>()
+        private val stack = mutableListOf<Cached>()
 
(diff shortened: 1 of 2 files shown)


spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/lifecycle/LifecycleAwareAdapter.kt
```diff
@@ -9,28 +9,29 @@ import kotlin.reflect.KProperty
 
 sealed class LifecycleAwareAdapter<T>(val factory: () -> T, val destructor: (T) -> Unit)
     : LifecycleAware<T>, LifecycleListener {
-    var cached: T? = null
+    protected data class CachedValue<out T>(val value: T)
```
artem-zinnatullin: Why protected?

artem-zinnatullin: Let's make a sealed class instead, so we could have `object` as null value thus we won't need `cached!!.value`

artem-zinnatullin (changes requested):

spek-runtime/common/src/main/kotlin/org/spekframework/spek2/runtime/lifecycle/LifecycleAwareAdapter.kt
tokou: * If you mean "why protected instead of private ?", it's because it can't be seen by subclasses if set to private (namely `ScopeCachingModeAdapter`, `TestCachingModeAdapter`, and `GroupCachingModeAdapter`)
* If you mean "why add a visibility modifier ?", then I ask you why should it be world visible ? I don't think that the property is meant to be read or written by other classes. Isn't it. Maybe I'm not understanding correctly the usage of these classes.

OK I'll give it a try with a `sealed class` instead. If I understand correctly it would be something like this ?
```
sealed class CachedValue {
    object Empty : CachedValue()
    data class Value<out T>(val value: T) : CachedValue()
}
```

artem-zinnatullin: - Ah ok, didn't realize it applies for classes in same file :)
- nono, I was like why not "private" :)
- yep!

tokou:
Gave it a try with sealed class.
Not sure I like it over the other solution though as type erasure makes it that we have to explicitly cast the value type to the generic type even in pattern matching structure. Used an extension function to hide that but still a bit ugly.

Your call on which solution to keep or if you have some good ways to make it less ugly

artem-zinnatullin:
To deal with type erasure you need to declare generic on sealed class itself :)

Check https://github.com/gojuno/koptional/blob/63be2777c929a1ddec6dc45744b572711616add6/koptional/src/main/kotlin/com/gojuno/koptional/Optional.kt#L3

tokou:
Ha! Thanks, thought about that but it didn't occur to me to make `Empty` a `Cached<Nothing>`

Still have a problem though, we still need to cast as `cached` is a `var` so the compiler is unable to do the smart cast :/

Maybe there's something else I'm missing ?

artem-zinnatullin:
Hm, either we need to leave classcast or we need to make local copy of the property value

I'd probably have local copy, since state when it was non-empty and then became empty is impossible even in multithreaded test

artem-zinnatullin (approved):
great!

artem-zinnatullin:
Thanks for your contributions, Tarek :)

Issue #371, opened 2018-02-26
Fail if a Test contains another Test


robstoll:
I accidentally wrapped a test into another test and only realised it because I used TDD and the Test did not fail as expected. I am not using TDD all the time and would like to be informed of such a mistake by Spek. So, the test `it has a test in it` should fail at the end because it contained another test.
```
object Spec: Spek({
    it("has a test in it") {
        test("foo") {
            throw IllegalStateException("will never be shown")
        }
    }
})
```

raniejade:
Related to #331, using `@DslMarker` should fix this as well.

raniejade:
Resolved by #398

PR #383, opened 2018-03-26
Prevent nested tests and actions

### LINKED ISSUE #331: Prevent users from nesting actions.
```kotlin
Spec : Spek({
  on("some action") {
  
    action.run()

    it("does something") {}

    on("nested action") {
      // We should either support that or fail the spec, ideally at compile time.
    }
  }
})
```

Possible solutions:

- Detect problem at runtime and throw exception
- `@DslMarker` https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-dsl-marker/

tokou:
Fixes #331 and #371

### DIFF
--- a/spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/ActionBody.kt
+++ b/spek-dsl/common/src/main/kotlin/org/spekframework/spek2/dsl/ActionBody.kt
@@ -24,4 +24,19 @@ interface ActionBody: TestContainer {
     fun afterGroup(callback: () -> Unit) {
         throw UnsupportedOperationException()
     }
+
+    @Deprecated(message = "Can't be used within action scopes", level = DeprecationLevel.ERROR)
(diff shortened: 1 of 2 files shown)


artem-zinnatullin:
Tests?

tokou:
I have a test that fails the compilation of the project.
But don't know how to write it an assertion about that, that's why I didn't submit it :)

tokou:
```
@Test
    fun testNestedSpek() {
        class TestSpek: Spek({
            group("group") {
                test("test") {
                    test("another test") {}
                }
                action("action") {
                    action("another action") {}
                }
            }
        })
    }
```

For reference

artem-zinnatullin:
Well, yeah `DeprecationLevel.ERROR` prevents compilation

Ok, leave it for us then, we can add integration test later :)

artem-zinnatullin (approved):
LGTM :)

raniejade:
This is now obsolete since we now use `@DslMarker` to prevent nesting. Sorry for not merging this asap.

PR #384, opened 2018-03-27
[FIX 1.x] Spek without default constructor

CedrickFlocon:
Hi,

For a generic setup we use a spek class without a default constructor
```kotlin
class Foo<T>(val bar: Any) : Spek({
    beforeEachTest {
        bar.startSomething()
    }

    afterEachTest {
        bar.finishSomething()
    }
})
```
And then use it like this
```kotlin
object MySpek : Spek({

    include(Foo(bar))

    ...

})
```
I am not sur this is the good way to generify some setup accros a lot of spek ?
By the way the `SpekTestEngine` will crash when it will try to create the spek test instance.

On this pull request a spek should have a default constructor.
I think it's not what we want because we can use `org.jetbrains.spek.api.CreateWith` annotation to let the `SpekTestEngine` know how to instantiate the spek (without a default constructor).
To be more robust I think we could change the `InstanceFactory` and let `create` method return a nullable.
```
interface InstanceFactory {
    fun <T: Spek> create(spek: KClass<T>): T?
}
```
Then the `defaultInstanceFactory` will return null if the spek don't have a default constructor.
At the end we just need to check if the spek instance is null before `invoke` it.

If you would like the second option just tell me I will update my pull request.

Best


raniejade:
You can mark it as `abstract` , unfortunately `1.x` is on maintenance mode so we won't merge anything to it unless it's critical.

PR #387, opened 2018-04-05
Add another easier way to install the spek engine

guenhter:
Gradle 4.6 provides built-in support for JUnit5. This makes it also easier to install engines.

https://docs.gradle.org/4.6/release-notes.html


artem-zinnatullin (approved):
lgtm

artem-zinnatullin:
@raniejade @ming13 I'd actually just drop Gradle <= 4.6 instructions, wdyt?

raniejade:
I'm good dropping it. 👍

PR #392, opened 2018-04-16
Introduce Specification based DSL

### LINKED ISSUE #374: Split Specification and Gherkin-style DSLs
The [Spectrum](https://github.com/greghaskins/spectrum) BDD testing library has an interesting approach to the `describe` + `context` vs. `given` + `when` dilemma. Basically [it identifies them as separate styles](https://github.com/greghaskins/spectrum/tree/1.2.0/docs#supported-features) and hides DSL in different packages.

Personally I think it is an awesome idea. Unfortunately I have no good ideas about how we can adopt this besides making it extra-clear in docs.

@raniejade @artem-zinnatullin What do you think?

raniejade:
DSL based on jasmine.

Played around with how to prevent users mixing two styles, my current solution is to create another `Spek` class that lives under `org.spekframework.spek2.specification`, which is basically a subclass of `org.spekframework.spek2.Spek`. It adds a two new receiver types: `Specification` which wraps `org.spekframework.spek2.dsl.Spec` and `SpecSuite` which wraps `org.spekframework.spek2.dsl.SpecBody`. I say _wrap_ because those receiver types don't extend their equivalent type. I did this to prevent leaking the base API.

Users will only need to depend on`org.spekframework.spek2:spek-dsl-specification` (or `org.spekframework.spek2:spek-dsl-gherkin`).

This partly (and hopefully) resolves #374.

cc @hhariri


raniejade:
Also tempted to do something similar to https://github.com/cucumber/cucumber/wiki/Gherkin 😂

artem-zinnatullin:
Hm, I need more time to look at this

raniejade:
Please focus on how I isolated the DSL :)

raniejade:
@ming13 if you could also take a look 🙏

raniejade:
One problem with this approach is that it will be hard to share code between different DSLs (plugins for example) since we are wrapping the base DSL. One problem with using extensions is you can still mix in different styles (even if you separate them into different artifacts).

arturdryomov:
I think the `context` is missing here.

Not sure if it is the right thing to do.

* Maybe the better approach is actually to use extensions in different packages to separate styles. That’s how Spectrum does it (at least). It brings some usability painpoints, i. e. you should import statements manually, but oh well.
* Another one can be borrowed from Scala [`FeatureSpec`](http://doc.scalatest.org/1.8/org/scalatest/FeatureSpec.html), but the current Spek classes organization prevents us from that.

I wonder if anyone else provides two styles to write tests. RSpec goes with `describe` and `context` only, the same as Quick for Swift.

I’m still gonna advocate everyone I see against using `Action` and `on` though ;-)

raniejade:
> Another one can be borrowed from Scala FeatureSpec, but the current Spek classes organization prevents us from that.

WDYM? 

> I’m still gonna advocate everyone I see against using Action and on though ;-)
That's why I tried to wrap action within a DSL, it will be useful for to do a DSL similar to gherkin.

arturdryomov:
@raniejade, well the current structure is something like this.

```kotlin
class ClassSpec : Spek({
    describe("things") {
        it("works") { assertTrue() }
    }
})
```

Scala uses traits for that, for example [`GivenWhenThen`](http://doc.scalatest.org/1.8/org/scalatest/GivenWhenThen.html). We can use `interface` with default implementation but that means passing instructions to some global state machine. I think Spectrum works this way — the test class is being created, then all instructions in constructor are getting executed and you have your test tree this way. Spek actually executes methods on `SpekBody`. In other words, default implementation methods operating on some _static_ state would not work. Sorry if I’m terribly wrong.

raniejade:
Kotlin doesn't have traits though, and not sure if we could ever emulate it. I think Spectrum works the same as Spek, it constructs the test tree when the runner is created (https://github.com/greghaskins/spectrum/blob/master/src/main/java/com/greghaskins/spectrum/Spectrum.java#L251). Then execution happens when JUnit decides to execute the runner. 

Anyway, one thing we could do is to only have one entry point for each style.

```kotlin
fun SpecBody.describe(description: String, body: SpecificationBody.() -> Unit) {
   ....
}

fun SpecBody.feature(description: String, body: FeatureBody.() -> Unit) {
    ...
}

```

Under the hood, we will be using the scope methods (I feel like we need to come up with a better name for them): `group`, `action`, and `test`. A more extreme change would be introducing a subclass for each style, to completely isolate the two DSLs.

```kotlin
class MySpec: Specification({
   ...
})

class MyFeature: Feature({
   ...
})
```

With the new IJ plugin, if you properly use `@Synonym` you can get IDE support pretty much automatically.

arturdryomov:
Let’s hear other people out then and meanwhile maybe document the desired API for each style?

Scala Test has [a nice page](http://www.scalatest.org/user_guide/selecting_a_style) enumerating everything they have.

raniejade:
Sure, I'll move the discussion to #374.

PR #408, opened 2018-05-07
Introduce Specification style

### LINKED ISSUE #394: Introduce specification style (jasmine, rspec)
Isolation is discussed on #374 , while this issue will discuss one of the styles we want to support. This style is  inspired by test frameworks like Jasmine and RSpec. 

```kotlin
class SetSpec: Spek({
    describe("a set") {
        val mySet by memoized { ... }
        
        context("when empty") {
            it("should have a size of 0") { ... }
            it("should throw an NoSuchElementException when ...") { ... }
        }
    }
})
```

This is equivalent to(using only Spek core scopes):

```kotlin
class SetSpec: Spek({
    group("describe a set") {
        val mySet by memoized { ... }
        
        group(" context when empty") {
            test("it should have a size of 0") { ... }
            test("it should throw an NoSuchElementException when ...") { ... }
        }
    }
})
```
###  Keywords

- `it` creates a spec, in JUnit world this is your test method. 
- `describe` creates a suite, which is a collection of specs. 
- `context` is an alias of `describe`.

### Nesting
Suites can be arbitrary nested, which allows you structure your specs.

### Pending
Each keyword will have a variant prefixed with an `x`. Pending suites and specs will show up in the execution tree but won't be actually executed.

raniejade:
Resolves #394.

It also does a few things:

- drop `samples`, it doesn't add any value anyway.
- rework several tests and extensions to only use the base DSL (group, action and test).


phatblat:
Nice!

PR #419, opened 2018-06-10
Remove @since from documentation.

arturdryomov:
It makes sense to have these in situations when compile version differs from the runtime version. I. e. JDK, Android SDK. Fortunately we are not in this state. API changes should be documented in releases anyway.


raniejade (approved):
Thanks! Been wanting to remove that for a long time!

PR #424, opened 2018-06-11
Fix snapshot builds.

arturdryomov:


raniejade (approved):

raniejade:
Snapshots don't include the IJ plugins, we have to manually publish them. A local android studio is required to build the AS plugin. If you know a way we can install it in travis - pleasee 🙏 .

arturdryomov:
@raniejade, it certainly can be automated and we are in luck — @artem-zinnatullin currently works on automating headless IJ. Not in the same way though.

Our friends at EL Passion develop [a Mainframer plugin](https://github.com/elpassion/mainframer-intellij-plugin), you can take a look at their configuration to find something useful.

vanniktech (approved):

artem-zinnatullin:
@raniejade if you could explain what needs to be done with AS I can try to automate that :)

PR #427, opened 2018-06-12
Rename Pending metadata to Skip.

arturdryomov:
I was totally confused by the naming when reading the code since the name doesn’t quite match the meaning. It is used for exclusion, i. e. for `xdescribe` and `xcontext`.

The naming does not align with [the RSpec terms](https://relishapp.com/rspec/rspec-core/v/3-7/docs/pending-and-skipped-examples) as well.

> An example can either be marked as skipped, in which is it not executed, or pending in which it is executed but failure will not cause a failure of the entire suite.


arturdryomov:
@raniejade, 🏓

raniejade (approved):

PR #428, opened 2018-06-13
Remove experimental annotation from non-experimental classes.

arturdryomov:
The only remaining `@Experimental` consumer is `InstanceFactory`. Really not sure if we need it at all since I don’t see a lot of usage here. It is kind of superseded by `Root.registerListener` from my  point of view. WDYT?

Also moved `@Ignore` to `.meta`, but we need to rethink the packaging a bit (or not).


artem-zinnatullin (approved):

spek-dsl/common/src/main/kotlin/org/spekframework/spek2/style/gherkin/Gherkin.kt
```diff
@@ -30,7 +30,7 @@ class Scenario(delegate: ActionBody) : ActionBody by delegate {
     }
 
     fun Then(description: String, body: TestBody.() -> Unit) {
-        test("Then $description", body = body)
+        test("And $description", body = body)
```
raniejade: Accidental change?

raniejade (changes requested):

spek-dsl/common/src/main/kotlin/org/spekframework/spek2/style/gherkin/Gherkin.kt
arturdryomov: Yep! Thanks, removed it.

raniejade (approved):
`InstanceFactory`is used in cases you want to control the classloader used to run the tests.

arturdryomov:
> InstanceFactoryis used in cases you want to control the classloader used to run the tests.

@raniejade, yep, but in which cases do you want to control that?

raniejade:
Robolectric tests for example.

PR #431, opened 2018-06-15
Remove action().

### LINKED ISSUE #422: Remove Action in favour of Group and Test combination
### Issue

At this point there are two ways to do the same thing — using `action` and `group`.

```kotlin
group("service") {
    
    beforeEachTest { 
        service.start()
    }
    
    test("it calls backend") {
        verify(backend).call()
    }
}
```
```kotlin
action("service") {

    service.start()
    
    test("it calls backend") {
        verify(backend).call()
    }
}
```
The difference is obvious — it is possible to omit `beforeEachTest` when using `action`. At the same time `action` cannot be nested. Before 2.0 this could lead to nasty bugs related to putting nested groups into `action` which were essentially ignored. Fortunately enough `@DslMarker` magic solved this for us. It leaves a couple of open issues though.

#### Alien Concept

Judging from experience — not a lot of people recognize what the difference is and when to use `group` and when to use `action`. And it is actually understandable. People coming from JUnit have to live with two completely new concepts (`action` and `group`) because the previous experience introduced them only to `test`. People coming from other languages do not get it as well — as far as I know RSpec from Ruby, Mamba from Python and Quick from Swift do not have `action` concept at all. The concept of `action` becomes kind of alien to them.

BTW — at current projects I work on we just ban `action` altogether to avoid bugs (2.0 is not released yet). And even 2.0 would not solve things like #388.

#### Scale

At the same time `action` does not scale well. I’ve encountered a lot of situations when product changes extend the behaviour. I. e. _click shows content_ becomes _click shows content and another click shows screen_. Fortunately we have nesting but `action` in such case needs to be migrated to `group` anyway since `action` does not support nesting.

#### Syntax

Since we are leaning to have two styles in 2.0 it becomes not quite obvious what is `action` and what is `group`.

* Specification — `describe` (`group`) + `context` (`group`) + `it` (`test`).
* Gherkin — `given` (`group`) + `when` (`group`) + `then` (`test`) + `and` (`test`).

There is basically no room for `action` at this point.

#### Tech

Internally having both `group` and `action` brings some complexity. I would not say it makes things drastically worse but it is something to have to live with.

---

CC @raniejade @hhariri @artem-zinnatullin

artem-zinnatullin:
This PR resolves #422 and removes Gherkin from Spek, PR #430 showcases how one could build Gherkin as an extension on top of Spek.


arturdryomov:
@artem-zinnatullin, please rebase, but only remove `Action`.

artem-zinnatullin:
Done

arturdryomov (approved):

vanniktech (approved):

raniejade (approved):

PR #432, opened 2018-06-18
Rework Gherkin style

raniejade:
So we can proceed with #422.

### DIFF
--- a/integration-test/common/src/test/kotlin/org/spekframework/spek2/integration/SetFeature.kt
+++ b/integration-test/common/src/test/kotlin/org/spekframework/spek2/integration/SetFeature.kt
@@ -1,20 +1,20 @@
 package org.spekframework.spek2.integration
 
 import org.spekframework.spek2.Spek
-import org.spekframework.spek2.style.gherkin.Feature
+import org.spekframework.spek2.style.gherkin.Given
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 
 object SetFeature: Spek({
-    Feature("Set") {
+    Given("Set") {
         val set by memoized { mutableSetOf<String>() }
 
-        Scenario("is empty") {
+        When("is empty") {
             Then("should have a size of 0") {
                 assertEquals(0, set.size)
             }
 
-            And("should throw when first is invoked") {
+            Then("should throw when first is invoked") {
                 assertFailsWith(NoSuchElementException::class) {
                     set.first()
                 }
--- a/spek-dsl/common/src/main/kotlin/org/spekframework/spek2/style/gherkin/Gherkin.kt
+++ b/spek-dsl/common/src/main/kotlin/org/spekframework/spek2/style/gherkin/Gherkin.kt
@@ -1,47 +1,40 @@
 package org.spekframework.spek2.style.gherkin
 
-import org.spekframework.spek2.dsl.ActionBody
 import org.spekframework.spek2.dsl.GroupBody
 import org.spekframework.spek2.dsl.Skip
 import org.spekframework.spek2.dsl.TestBody
 import org.spekframework.spek2.meta.*
 
 @SpekDsl
-class Feature(delegate: GroupBody) : GroupBody by delegate {
-    @Synonym(SynonymType.ACTION, prefix = "Scenario: ")
+class GivenBody(delegate: GroupBody) : GroupBody by delegate {
+    @Synonym(SynonymType.GROUP, prefix = "When: ")
     @Descriptions(Description(DescriptionLocation.VALUE_PARAMETER, 0))
-    fun Scenario(description: String, skip: Skip = Skip.No, body: Scenario.() -> Unit) {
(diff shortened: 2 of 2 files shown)


arturdryomov (approved):

integration-test/common/src/test/kotlin/org/spekframework/spek2/integration/SetFeature.kt
```diff
@@ -1,20 +1,20 @@
...
         val set by memoized { mutableSetOf<String>() }
 
-        Scenario("is empty") {
+        When("is empty") {
             Then("should have a size of 0") {
                 assertEquals(0, set.size)
             }
 
-            And("should throw when first is invoked") {
+            Then("should throw when first is invoked") {
```
artem-zinnatullin, line 17: Wait, so by default `memoized` is `TEST` scoped, doesn't this have to be nested now as @ming13 suggested?

raniejade: `Then` is a test. Not sure if I get what you're saying.

artem-zinnatullin: yes, but this `SetFeature` is actually stateless, if you try to add stateful test, you'll see that it doesn't play well with Gherkin because memoized resets state for every `Then`

unless we recommend to do changes in `When` using `beforeEachTest`

arturdryomov:
@raniejade, maybe let’s merge it and resolve issues further? I have other suggestions about DSL that actually might overlap with the current state 😉

integration-test/common/src/test/kotlin/org/spekframework/spek2/integration/SetFeature.kt
NorseDreki: (Not sure I'm completely following these changes, but)

>if you try to add stateful test, you'll see that it doesn't play well with Gherkin because memoized resets state for every Then

According to Gherkin, state should be shared across `Then`s, for example:

>There are examples when it not possible to split a scenario. Withdrawing money from a bank account is such an example.
The customer should get their money from an ATM machine, say £50
The customer account should be debited £50

(from http://www.thinkcode.se/blog/2016/06/22/cucumber-antipatterns)

So, for one `When` there are two `Then`s performing checks on a shared state.
And I'd say it's a pretty common situation.

NorseDreki:
I'd vote that bringing in Gherkin-like style adds to an existing amount of confusion with "given-on-it" / "describe-context-it" styles.

From own experience:
- `given` hinted me it was Gherkin and it took time to realize it was not;
- nesting for "given-on-it" requires using `describe` / `context` from another style (unless one decides to go weird like "given-given-on-it"), thus styles are mixed;

So maybe proper implementation of Gherkin as an add-on (as was suggested elsewhere) would simplify lives of newcomers, otherwise this library (which I like a lot) might end up having a multitude of hardly distinguishable styles (as here, for example: https://github.com/kotlintest/kotlintest/blob/master/doc/styles.md).

IMHO.

raniejade:
@ming13 @artem-zinnatullin should we allow test scopes to have children? They shouldn't be treated as scopes but more of a way to group assertions/actions. 

```kotlin
test("Given: Something") {
  val foo by memoized { ... }

  something("When: this") { ... }
  something("Then: that") { ... }
}
```

Using the proper synonyms:

```kotlin
Given("Something") {
  val foo by memoized { ... }

  When("this") { ... }
  Then("that") { ... }
}
```

arturdryomov:
Seems like `Action` all over again :-) Not sure if that can be useful.

raniejade:
But without the extra baggage. Just treat as a way to build much richer DSLs. There's no lifecycle attached to it, just a way to add a child to a test in the test tree.

arturdryomov:
But why do you need a test inside of a test? It is like

```kotlin
@Test fun outer() {
    @Test fun inner() {
        @Test fun innerInner() {
        }
    }
}
```

I think it discourages using proper grouping this way.

raniejade:
It's still one test, the only difference is the test tree can contain much more fine grained steps.

arturdryomov:
Isn’t it the whole point of test to have a single check?

raniejade:
But you have setup, action and the check as well within the test. Is there anything wrong to make it much more apparent?

arturdryomov:
Not sure what real world example do you have in mind but personally I have only checks inside of a test.

```kotlin
context("click button") {
    
    beforeEachTest {
        view.button.click()
    }

    it("shows message") {
        verify(view).showMessage()
    }
}
```

raniejade:
How would we support other styles? Gherkin for example. If we go this route, every style is the same just with a different synonym. What's the point of supporting other styles then?

raniejade:
I don't think it's bad to introduce it in the core dsl, it gives more flexibility to users who want to write their own style. It shouldn't affect the specification style, since it won't be using it.

PR #433, opened 2018-06-21
Remove execution context.

arturdryomov:
`ExecutionContext` was just a wrapper on top of `ExecutionRequest` and wasn’t actually used anywhere except `Executor`.

Had to remove a bunch of `Action`-related code @artem-zinnatullin forgot about.

### DIFF
--- a/spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/SpekTestDescriptor.kt
+++ b/spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/SpekTestDescriptor.kt
@@ -34,7 +34,6 @@ class SpekTestDescriptor internal constructor(
 
     override fun getType(): TestDescriptor.Type = when (scope) {
         is GroupScopeImpl -> TestDescriptor.Type.CONTAINER
-        is ActionScopeImpl -> TestDescriptor.Type.CONTAINER_AND_TEST
         is TestScopeImpl -> TestDescriptor.Type.TEST
     }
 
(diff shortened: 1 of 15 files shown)


vanniktech (approved):
> Had to remove a bunch of Action-related code @artem-zinnatullin forgot about.

Can happen to anyone :man_shrugging: we also reviewed and didn't catch it.

arturdryomov:
@vanniktech, yeah, that’s mostly a note why so many changes. I’ve removed a single thing and everything else just followed.

spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/SpekTestDescriptor.kt
```diff
@@ -70,7 +69,7 @@ class SpekTestDescriptor internal constructor(
 
     override fun getChildren() = childDescriptors
 
-    override fun mayRegisterTests(): Boolean = scope is ActionScopeImpl
+    override fun mayRegisterTests(): Boolean = false
```
raniejade: We can omit this since the default implementation returns `false`.

raniejade (changes requested):
One comment, then LGTM.

spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/SpekTestDescriptor.kt
arturdryomov: Nice! Done.

raniejade (approved):

PR #459, opened 2018-07-22
Fix class name lookup for jdk10

raniejade:


spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
```diff
@@ -59,7 +59,7 @@ actual class SpekRuntime : AbstractRuntime() {
 
     private fun createReflections(testDirs: List<String>): Reflections {
         val urls = if (testDirs.isEmpty()) {
-            ClasspathHelper.forClassLoader()
+            ClasspathHelper.forJavaClassPath()
```
raniejade, line 62: returns an empty list on java 10, using `java.class.path` system property instead.

spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/scope/Path.kt
```diff
@@ -83,9 +83,11 @@ actual class PathBuilder(private var parent: Path) {
         actual val ROOT: Path = Path("", null)
 
         actual fun from(clz: KClass<out Spek>): PathBuilder {
+            val packageName = clz.java.`package`.name
+            val className = clz.java.name.removePrefix("$packageName.")
             return PathBuilder()
-                .append(clz.java.`package`.name)
-                .append(clz.java.simpleName)
```
raniejade, line 88: `simpleName` on java 10 only returns the leaf name of an inner class i.e `TestSpek` instead of `AfterEachTest$testSomething$TestSpek` which is the case for java 8.

PR #463, opened 2018-08-01
Use DynamicTestExecutor to fix incompatibility

### LINKED ISSUE #462: Incompatibility with JUnit Platform 1.3-M1
Originally reported as https://github.com/junit-team/junit5/issues/1518.

Spek 1.1.5 seems to completely ignore the passed `DynamicTestExecutor` [1] and instead add the test descriptor and call the execution listener directly [2] which is not how `HierarchicalTestEngine` is supposed to be used. 🤔 

[1] https://github.com/spekframework/spek/blob/v1.1.5/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/Scope.kt#L44-L53

[2] https://github.com/spekframework/spek/blob/v1.1.5/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/SpekTestEngine.kt#L214-L215

@raniejade Is Spek 1.x still maintained? Would a PR be accepted?

marcphilipp:
Prior to this commit the dynamically discovered tests were reported to the `EngineExecutionListener` directly which caused an incompatibility with JUnit Platform 1.3.

However, merging this PR would change the runtime behavior. Previously, a scope block was executed entirely before executing the first test. Now, the tests are executed immediately when they are encountered.

Is this acceptable or do we need a different solution?

Fixes #462.

### DIFF
--- a/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/Scope.kt
+++ b/spek-junit-platform-engine/src/main/kotlin/org/jetbrains/spek/engine/Scope.kt
@@ -32,7 +32,7 @@ sealed class Scope(uniqueId: UniqueId, val pending: Pending, val source: TestSou
     class Action(uniqueId: UniqueId, pending: Pending,
                  source: TestSource?,
                  lifecycleManager: LifecycleManager,
-                 private val body: Action.(SpekExecutionContext) -> Unit)
+                 private val body: Action.(Node.DynamicTestExecutor) -> Unit)
         : Scope(uniqueId, pending, source, lifecycleManager), ActionScope {
         override fun getType() = CONTAINER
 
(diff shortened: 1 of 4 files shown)


marcphilipp:
If the runtime behavior change is not acceptable, I could envision JUnit adding an API like this:
https://github.com/junit-team/junit5/blob/64366f8a90e3a09f76d5f442fd716918ede418e4/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java#L246-L253

raniejade (approved):
@marcphilipp Tested locally, working as usual. Thank you! Just curious, code worked pre 1.3 - what did change?

marcphilipp:
To support parallel execution we now need to analyze the complete test tree for locks before executing it. Thus, we walk the whole tree once. Children that are added to `TestDescriptors` later will simply be ignored. Previously, they were executed after calling `execute()` for their container.

BTW JUnit Platform 1.3 now supports parallel execution. You can opt-in to that for Spek by overriding `createExecutorService` from `HierarchicalTestEngine` and return a `ForkJoinPoolHierarchicalTestExecutorService`, e.g. based on a config param. See the `JupiterTestEngine` for an example.

marcphilipp:
@raniejade Which version will this fix be released in?

raniejade:
Will be included in `1.2.1`, so hopefully by next week. I'll see how we could include parallel execution in `2.x`, does it support parallelizing only a specific container? i.e at a class level.

raniejade:
Also in `2.x`, we have completely decoupled from JUnit Platform. It's just a thin wrapper around a pure kotlin runtime, so we could support other kotlin platforms.

marcphilipp:
> does it support parallelizing only a specific container? i.e at a class level.

No it runs the whole tree in parallel.

> Also in 2.x, we have completely decoupled from JUnit Platform.

Yeah, I saw that. Unfortunately that means you cannot easily hook into the parallel execution support.

PR #471, opened 2018-08-16
Drop reflections in favor of classgraph

raniejade:
Looks like https://github.com/ronmamo/reflections is not actively developed. I have no numbers but classgraph should be faster as it doesn't depend on java runtime reflection. 

Reflections will also try to parse anything as a class: https://github.com/ronmamo/reflections/blob/master/src/main/java/org/reflections/Reflections.java#L251 hence the errors.

### DIFF
--- a/spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
+++ b/spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
@@ -1,9 +1,6 @@
 package org.spekframework.spek2.runtime
 
-import org.reflections.Reflections
-import org.reflections.scanners.SubTypesScanner
-import org.reflections.util.ClasspathHelper
-import org.reflections.util.ConfigurationBuilder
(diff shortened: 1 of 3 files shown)


spek-runtime/jvm/src/main/kotlin/org/spekframework/spek2/runtime/SpekJvmRuntime.kt
```diff
@@ -57,18 +52,20 @@ actual class SpekRuntime : AbstractRuntime() {
...
+            cg.overrideClasspath(testDirs)
         }
 
-        return Reflections(
-            ConfigurationBuilder()
-                .setUrls(urls)
-                .setScanners(SubTypesScanner())
-        )
+        return cg.scan().use {
+            val classes = it.getSubclasses(Spek::class.qualifiedName!!)
```
vanniktech: inline this one?

vanniktech (approved):

artem-zinnatullin (approved):
NICE!

vanniktech (approved):

Issue #475, opened 2018-08-21
Unclear test suite failures when using empty xdescribe blocks


mtrewartha:
I found an annoying scenario where the test suite fails to run when a spec contains an empty `xdescribe(...)` call. I often times write a skeleton of a test suite when TDDing some new code, so I wind up with a file like this:

```kotlin
object SomeNewClassSpec : Spek({
    xdescribe("some new functionality") {
    }
    xdescribe("some other new functionality") {
    }
})
```

I'll then progressively flesh the file out, implementing tests, then implementing code, running the tests as I go and watching them start to pass. However, when trying to do so, the suite fails to run and I get no feedback on the command line, but looking at the test report HTML, I see this:

```
org.gradle.api.internal.tasks.testing.TestSuiteExecutionException: Could not complete execution for Gradle Test Executor 22.
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:32)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:93)
	at com.sun.proxy.$Proxy2.stop(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.stop(TestWorker.java:131)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #476, opened 2018-08-22
Verify entire Spek can be skipped

artem-zinnatullin:
PR adds integration test that verifies that we don't crash execution if nothing was executed in the Spek due to `group(skip = Yes)`

Related to #475


vanniktech (approved):

artem-zinnatullin:
nvm, this test doesn't run through JUnit runner

PR #460, opened 2018-07-27
Build against Java 10

raniejade:

PR #477, opened 2018-08-22
Ignore unrecognized jvm options for Gradle daemon to unblock JDK8 builds

artem-zinnatullin:
PR #460 added jvm option for Gradle daemon that is not supported by JDK8, this PR suppresses default JVM behavior that causes build to fail with following message:

```console
./gradlew tasks
Starting a Gradle Daemon

FAILURE: Build failed with an exception.

* What went wrong:
Unable to start the daemon process.
This problem might be caused by incorrect configuration of the daemon.
For example, an unrecognized jvm option is used.
Please refer to the user guide chapter on the daemon at https://docs.gradle.org/4.10-rc-1/userguide/gradle_daemon.html
Please read the following process output to find out more:
-----------------------
Unrecognized option: --add-modules=java.xml.bind,java.activation
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

It's not a great solution, as it hides the fact that we can accidentally pass unsupported jvm option, but it unblocks JDK8 builds for now.

Ideally, we should be able to set different jvm options for different JDK versions

cc @vanniktech

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,4 +9,4 @@ systemProp.org.ajoberstar.grgit.auth.username=nothing
 systemProp.org.ajoberstar.grgit.auth.password=nothing
 ijSdkVersion=2018.2
 ijSdkKotlinPluginCoords=org.jetbrains.kotlin:1.2.60-release-IJ2018.2-1
-org.gradle.jvmargs=--add-modules=java.xml.bind,java.activation
+org.gradle.jvmargs=--add-modules=java.xml.bind,java.activation -XX:+IgnoreUnrecognizedVMOptions


raniejade (approved):
Should we also update travis job to build against JDK 8 as well?

vanniktech (approved):

artem-zinnatullin:
Well, as long as we target JDK 8+ at compile time which we do with Kotlin implicitly, we should be good

We can require project to be buildable on JDK 8 I guess, I don't have strong preference given how fast JDK now evolves

PR #478, opened 2018-08-26
Add kotlin-common module sample

raniejade:


samples/common/build.gradle
```diff
@@ -0,0 +1,25 @@
+buildscript {
+    ext.kotlin_version = '1.2.51'
```
vanniktech: Why aren't we using 1.2.60 here?

vanniktech (approved):

samples/common/build.gradle
raniejade: Oops, forgot to update that. Copy pasted from the multiplatform sample.

vanniktech (approved):

PR #481, opened 2018-09-01
Split AS plugin builds

raniejade:
Code-wise both plugins are the same but we need to create a separate plugin build for AS 3.1 and 3.2. IJ plugin versioning is not that great because it doesn't differentiate between android studio and the mainline products. Without this AS 3.2 will pull the plugin built for IJ which doesn't have android specific configurations.


spek-ide-plugin/android-studio3.2/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+}
+
+buildPlugin {
+    baseName = ''
+    version = "${project.version}-Studio3.1"
+}
+
+patchPluginXml {
+    version = "${project.version}-Studio3.1"
+    sinceBuild = "181.5281"
```
raniejade, line 25: `x.5281` here is a specific build number for AS 3.2.

vanniktech (approved):
I believe we as also need one for AS 3.3 then. If I recall correctly it is 183

raniejade:
Yep, I'll create a separate PR for that - need to download the canary builds to get the exact build numbers.

raniejade:
Wow, thanks @vanniktech!

PR #482, opened 2018-09-01
Update doc to use proper dependency

raniejade:


vanniktech (approved):
Although just using the sdtlib is also fine

raniejade:
@vanniktech Using only `kotlin-stdlib` is causing an error, complaining about a missing class related to Java 8 streams.

PR #493, opened 2018-09-16
Add gherkin docs

raniejade:


docs/gherkin.md
```diff
@@ -0,0 +1,72 @@
+This style brings in a Gherkin-like DSL into Spek.
```
vanniktech: dunno but I feel like `brings` is wrong. It's already in Spek isn't it? So this documentation does not bring it into it

vanniktech (approved):

docs/gherkin.md
raniejade: Oops, forgot to change that.

vanniktech (approved):

PR #497, opened 2018-09-20
Add preserveExecutionOrder flag

### LINKED ISSUE #496: Ensure execution order for the Gherkin DSL
Proposed solution is to add an optional param on the scope methods (`group` and `test`), which defaults to unspecified order. Can be an enum or just a boolean parameter. `Scenario` would set this variable to ensure execution when calling `group`.

see #494 for context

raniejade:
Resolves #496. Did not want to implement randomization yet, implementation remains as is but I've set the flag to `true` in the appropriate places. 

Not entirely happy with the parameter name, I'm open to suggestions. //cc @vanniktech @davidsowerby


vanniktech (approved):
I like the name :man_shrugging:

PR #499, opened 2018-09-23
Use virtualenv to build docs

raniejade:
This will allow us to use the latest version of mkdocs and mkdocs-material.


preview_docs.sh
```diff
@@ -0,0 +1,11 @@
+#! /usr/bin/env bash
+BASEDIR=$(dirname "$0")
```
artem-zinnatullin: please add `set -e` so that script fails if some command fails, otherwise it'll continue executing

artem-zinnatullin (approved):

build_docs.sh
```diff
@@ -0,0 +1,11 @@
+#! /usr/bin/env bash
+BASEDIR=$(dirname "$0")
```
vanniktech: Same here

vanniktech (approved):

vanniktech (approved):
Out of curiosity why always `.` as a commit message? :dancer:

raniejade:
@vanniktech Less keys to press than `fix something`, the whole PR gets squashed into one commit anyway.

vanniktech:
Yeah thought so.

PR #500, opened 2018-10-02
Add missing $ in Maven example

kdkeyser:
Property names need to be prefixed with $ to be replaced by their value. This was missing in the doc, so the resulting Maven example would not work


raniejade:
Thanks!

PR #504, opened 2018-10-04
Resolve Gradle deprecation warning

marcphilipp:
This PR enables the 'STABLE_PUBLISHING' feature preview which will become mandatory in Gradle 5.0. It changes the `publishing` extension to behave like any other Gradle extensions. Since its evaluation is no longer deferred, the `distribution` subproject needs to explicitly postpone configuring its publications until the corresponding subprojects have been evaluated. Otherwise, references to tasks and the `java` component won't work. As a bonus, the publication names can now be reused at configuration time to configure the `bintray` extension without having to duplicate them.

### DIFF
--- a/distribution/build.gradle
+++ b/distribution/build.gradle
@@ -74,7 +74,7 @@ def artifact(Publication publication, Project project, artifactName) {
 
     configure.resolveStrategy =  Closure.DELEGATE_FIRST
     configure.delegate = publication
-    configure()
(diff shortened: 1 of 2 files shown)


artem-zinnatullin (approved):

artem-zinnatullin:
This is really nice of you guys to fix oss projects that use Gradle, ❤️

vanniktech (approved):

PR #533, opened 2018-11-05
Rework IJ plugin introspection

### LINKED ISSUE #84: IntelliJ IDEA support for jumping to failing "it" block
The IDE support for showing test results in a given > on > it tree is great, it would be cool if double-clicking on a failing line here took you to the failing code block in question.

raniejade:
Rebuilds how the IJ plugin introspect if a source element is runnable. It now builds a tree representation of a test (and caches the value - updating when necessary). Any query for a source element is made against that tree (unlike before) - making it less expensive. Big bonus for this refactoring is navigating to source is now fully supported (resolves #84)!


vanniktech:
Let me know once this is done and I'll be happy to review it!

raniejade:
Yep, will let you know. Just fixing some performance issues then it should be good for a review!

raniejade:
@vanniktech ready for review. Few caveats tho:
- Evicting is a bit of a problem, I'll have to think of someway to do it - which I'm going to address in a separate PR. Shouldn't be a problem unless you have several thousand tests classes.
- navigating to source does not work with the `gherkin` style, I know the problem but fixing it is beyond the scope of this PR.

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/SpekRunConfigurationProducer.kt
```diff
@@ -31,14 +35,44 @@ abstract class SpekRunConfigurationProducer(val producerType: ProducerType, conf
 ) {
     override fun isConfigurationFromContext(configuration: SpekBaseRunConfiguration<*>,
                                             context: ConfigurationContext): Boolean {
-        val path = context.psiLocation?.let { extractPath(it, true) }
-        return configuration.path == path
+        val descriptor = context.psiLocation?.let {
+            val elementContext = PsiTreeUtil.getContextOfType(it, KtClassOrObject::class.java, KtCallExpression::class.java)
+            if (elementContext != null) {
```
vanniktech: If you're doing the `is` checks below you don't need to check for `!= null` and hence reduce the number of depth

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/SpekRunConfigurationProducer.kt
```diff
@@ -31,14 +35,44 @@ abstract class SpekRunConfigurationProducer(val producerType: ProducerType, conf
...
+                }
+            } else {
+                null
+            }
+        }
+
+        if (descriptor == null) {
+            return false
+        }
+        return configuration.path == descriptor.path
```
vanniktech: just `return configuration.path == descriptor?.path`? That way we can get rid of the if statement above

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/SpekRunConfigurationProducer.kt
```diff
@@ -31,14 +35,44 @@ abstract class SpekRunConfigurationProducer(val producerType: ProducerType, conf
...
     }
 
     override fun setupConfigurationFromContext(configuration: SpekBaseRunConfiguration<*>,
                                                context: ConfigurationContext,
                                                sourceElement: Ref<PsiElement>): Boolean {
-        val path = sourceElement.get().let { extractPath(it, true) }
+        val path = sourceElement.get().let {
+            val elementContext = PsiTreeUtil.getContextOfType(it, KtClassOrObject::class.java, KtCallExpression::class.java)
+            if (elementContext != null) {
```
vanniktech: same as above

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/SpekRunLineMarkerContributor.kt
```diff
@@ -4,27 +4,34 @@ import com.intellij.execution.lineMarker.ExecutorAction
...
-                }
-            }
-            else -> null
+        if (element !is KtNameReferenceExpression) {
+            return null
         }
 
+        val context = PsiTreeUtil.getContextOfType(element,false, KtClassOrObject::class.java, KtCallExpression::class.java)
+
+        val path = if (context != null) {
```
vanniktech: same as above

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/SpekRunLineMarkerContributor.kt
```diff
@@ -4,27 +4,34 @@ import com.intellij.execution.lineMarker.ExecutorAction
...
-                    is KtClassOrObject -> extractPath(parent)
-                    else -> null
-                }
-            }
-            else -> null
+        if (element !is KtNameReferenceExpression) {
+            return null
         }
 
+        val context = PsiTreeUtil.getContextOfType(element,false, KtClassOrObject::class.java, KtCallExpression::class.java)
```
vanniktech: nit: missing a space after `element,`

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/base.kt
```diff
@@ -92,12 +99,29 @@ abstract class SpekBaseRunConfiguration<T: RunConfigurationModule>(name: String,
...
             "$prefix Spek tests ${path.name}"
         } else {
             "$prefix ${path.name}"
         }.trim()
     }
 }
+
+object SpekScopeLocator: SMTestLocator {
+    override fun getLocation(protocol: String, path: String, project: Project, scope: GlobalSearchScope): MutableList<Location<PsiElement>> {
+        if (protocol != "spek") {
```
vanniktech, line 117: Can this ever the be case? If not we could even throw here

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/base.kt
```diff
@@ -92,12 +99,29 @@ abstract class SpekBaseRunConfiguration<T: RunConfigurationModule>(name: String,
...
+            return mutableListOf()
+        }
+        val locations = mutableListOf<Location<PsiElement>>()
+        val path = PathBuilder.parse(path).build()
+        val descriptor = ScopeDescriptorCache.findDescriptor(path)
+
+        if (descriptor != null) {
+            locations.add(PsiLocation(descriptor.element))
```
vanniktech: not a fan of MutableLists. Would just change it to

```kotlin
return if (descriptor != null) listOf(PsiLocation(descriptor.element)) else emptyList
```

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
```diff
@@ -0,0 +1,226 @@
...
+    class Group(path: Path, element: KtElement, val children: List<ScopeDescriptor>): ScopeDescriptor(path, element) {
+        fun findDescriptorForElement(element: KtElement): ScopeDescriptor? {
+            return findMatching(this, element)
+        }
+
+        private fun findMatching(current: ScopeDescriptor, element: KtElement): ScopeDescriptor? {
+            if (current.element == element) {
+                return current
+            } else if (current is ScopeDescriptor.Group) {
+                for (child in current.children) {
```
vanniktech: We could use `current.children.find { findMatching(it, element) }`

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
```diff
@@ -0,0 +1,226 @@
...
+        return ScopeDescriptor.Group(path, clz, children.toList()).apply {
+            index[path.serialize()] = this
+        }
+    }
+
+    private fun buildScopes(parent: Path, block: KtBlockExpression): List<ScopeDescriptor> {
+        val callExpressions = PsiTreeUtil.getChildrenOfTypeAsList(block, KtCallExpression::class.java)
+        val scopes = mutableListOf<ScopeDescriptor>()
+
+        callExpressions.forEach { callExpression ->
```
vanniktech: I'd turn this into a stream and then use `filter()` `mapNotNull` `flatMap` etc as I find it very hard to read this way since it's a mixture of both

vanniktech (approved):
Still don't grasp all of the internals so I was mostly looking out for other things

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/base.kt
artem-zinnatullin: yis pls

artem-zinnatullin: also avoids allocation in empty case that way

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
```diff
@@ -0,0 +1,226 @@
...
+
+private val SYNONYM_CLASSES = listOf(
+    "org.spekframework.spek2.meta.Synonym"
+)
+
+private val DESCRIPTIONS_CLASSES = listOf(
+    "org.spekframework.spek2.meta.Descriptions"
+)
+
+object ScopeDescriptorCache {
```
artem-zinnatullin: Can we make this non-static? I'm afraid that over time there'll be some awkward usages of this cache because it's global static variable

artem-zinnatullin: also maybe extract into its own file

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
```diff
@@ -0,0 +1,226 @@
+package org.spekframework.intellij.domain
```
artem-zinnatullin: this is a lot of code that seems easy to make a mistake in and hard to change in future, do you think there are reasonable ways to unit-test it?

Maybe something similar to [Android Lint tests](https://github.com/googlesamples/android-custom-lint-rules/blob/439605d4b10afa87cb58ba500d4c4b9a219770bf/android-studio-3/checks/src/test/java/com/example/lint/checks/SampleCodeDetectorTest.java) when you pipe source to it and run UAST/PSI analysis as a test

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/base.kt
raniejade: TIL: `emptyList()` 👍

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
raniejade: 👍

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
raniejade: Till today I still don't know how to test IJ plugins 😂 . But yeah, I'll add in some tests.

vanniktech: No one does!

spek-ide-plugin/intellij-common/src/main/kotlin/org/spekframework/intellij/domain/descriptors.kt
raniejade: I need the result of `findMatching` not the current child.

raniejade:
changes are in (mostly). I'll defer the test to another round of coding, the tests require a proper project setup (so it can see the @Synonym annotations)  which I currently don't know how to do.

vanniktech (approved):

artem-zinnatullin (approved):

PR #534, opened 2018-11-07
Update samples to use latest version

raniejade:


samples/android/build.gradle
```diff
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.2.60'
+    ext.kotlin_version = '1.2.71'
```
vanniktech: There's 1.3.0

samples/android/build.gradle
```diff
@@ -19,8 +19,6 @@ buildscript {
 allprojects {
     repositories {
         google()
-        mavenCentral()
         jcenter()
-        maven { url "https://dl.bintray.com/spekframework/spek-dev" }
```
vanniktech, line 24: We don't need that anymore?

vanniktech (approved):

artem-zinnatullin (approved):

samples/android/build.gradle
raniejade: Instead of removing it, I'll just add a comment that it's required if you need a bleeding edge version

vanniktech (approved):

PR #538, opened 2018-11-19
Add runnable flag to Synonym annotation

raniejade:
Primary purpose is to allow IJ plugin to discover scopes that can't be run individually (`Given`, `When`, `Then` and `And`). This is needed so `Navigate to source` works with the gherkin style.


vanniktech (approved):

artem-zinnatullin (approved):
👍

PR #541, opened 2018-11-28
Spek1 over Spek2

raniejade:
Implemented for the lolz - not to be merged. Spek 1 implemented (partly) on top of Spek 2. Several caveats.

- Use `org.spekframework.spek2.Spek` instead of Spek 1's equivalent: `org.jetbrains.spek.api.Spek`.
- No subject support use `val subject by memoized { ... }`.
- No data-driven support.
- Not tested, possibly riddled with 🐛 .


vanniktech:
What's the motivation behind this?

PR #542, opened 2018-11-29
Fix run marker showing up for method call

raniejade:
Fix is not getting the **context** of a method call. 

```
it("should do something") {
   set.add(1)
}
```

The context of `set.add(1)` is  enclosing scope (`it(...)`)


vanniktech (commented):
:+1:

PR #548, opened 2018-12-15
Avoid doing child recalculation for cached junit descriptor

raniejade:
We cache the value but re-run the child calculation every time regardless if we hit the cache or not.

This test is fast when running within IJ but super slow when running via JUnit 5.

```kotlin
object LargeTest: Spek({
    repeat(27000) {
        test("test $it") {

        }
    }
})
```


vanniktech (commented):
:+1:

Issue #563, opened 2019-01-11
NoClassDefFoundError when using mocks and running spek from IntelliJ on a Windows machine


jcornaz:
Hello, 

Please consider the following test (using [MockK](https://mockk.io/)):

```kotlin
class MockkSpec : Spek({
  val mock = mockk<Greeter>()

  beforeEachTest {
    every { mock.greet(any()) } answers { "Hello ${args[0]}" }
  }

  afterEachTest {
    clearMocks(mock)
  }

  describe("greet(\"mock\")") {
    val result by memoized { mock.greet("mock") }

    it("should return \"Hello mock\"") {
      assertEquals("Hello mock", result)
    }
  }
})
```

It runs fine from gradle (`graldew test`). But running the test from IntelliJ fails with: 

```
Exception in thread "main" java.lang.NoClassDefFoundError: io/mockk/proxy/jvm/dispatcher/JvmMockKDispatcher
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:801)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:699)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:622)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:580)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:801)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:699)
[...]

PR #577, opened 2019-01-24
Fix Working Directory and JRE Fields in Plugin UI

floralvikings:
Previously the `data.workingDirectory` field was set to itself instead of the intended value, meaning it was wiped every time the run configuration was modified; now it is set correctly.

Similar story with the `alternativeJrePath` properties.  Found these while researching #563.


raniejade:
@floralvikings thanks! Can you apply the same changes to the android plugin? https://github.com/spekframework/spek/blob/2.x/spek-ide-plugin/android-studio/src/main/kotlin/org/spekframework/intellij/SpekAndroidSettingsEditor.kt

floralvikings:
Sure; I'll get it done this evening 👍

floralvikings:
@raniejade added changes to the android plugin 😄

raniejade:
Sweet, thanks again @floralvikings! Taking a look tonight!

PR #591, opened 2019-02-01
Upgrade to Kotlin 1.3.20

charleskorn:
Resolves https://github.com/spekframework/spek/issues/590.

I've switched to the simpler syntax for referring to Kotlin dependencies (eg. stdlib), but haven't touched anything in the IntelliJ plugins, as they're using the non-MPP Gradle plugin.


raniejade (approved):

raniejade:
thanks @charleskorn !

PR #594, opened 2019-02-02
Don't try to release from forks

### LINKED ISSUE #593: Don't trigger release builds on forks
Travis does not support it out of the box, but we could properly wrap the release commands in a shell script  then return early if the credentials are not present (`BINTRAY_API_KEY`, `BINTRAY_USER` and `HUB_API_TOKEN`).

charleskorn:
This fixes https://github.com/spekframework/spek/issues/593.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,9 +58,9 @@ jobs:
 stages:
   - build
   - name: release
-    if: tag IS present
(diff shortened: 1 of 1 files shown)


raniejade (approved):
TIL!

Issue #600, opened 2019-02-04
Implement console launcher for Kotlin Native


raniejade:
Relates to #555 

For prior art see: https://github.com/spekframework/spek/blob/2.x/spek-ide-plugin/interop-jvm/src/main/kotlin/org/spekframework/ide/console.kt.

I'm imagining something like:

```kotlin
interface ConsoleLauncher {
  fun launch(context: DiscoveryContext, args: List<String>): Int
}
```

then KN user's in the meantime can write:

```kotlin
fun main(args: List<String>) {
   // build discovery context
   val discoveryContext = DiscoveryContext.builder()
      ...
      .build()

  val consoleLauncher = KNConsoleLauncher()
  val exitCode = consoleLauncher.launch(discoveryContext, args)
  exitProcess(exitCode)
}
```

Within gradle, we can invoke something like:

```
<executable> --reporter=xunit --reports-dir={project-build-dir}/spek/reports
```

For IJ integration (eventually), we can issue a command like (`sm` is service message, see https://github.com/spekframework/spek/blob/2.x/spek-ide-plugin/interop-jvm/src/main/kotlin/org/spekframework/ide/ServiceMessageAdapter.kt):
```
<executable> --path <path-to-run> --reporter=console --console-reporter-format=sm
```

The actual reporter implementations can be done separately, this issue only covers the necessary infrastructure/api to write custom reporters.

Issue #599, opened 2019-02-04
Setup CI builds for Kotlin Native


raniejade:
Relates to #555.

The current travis builds only runs on linux, we'd want to build for windows and osx as well.

raniejade:
https://azure.microsoft.com/en-au/services/devops/pipelines/ looks interesting, going to try it out.

Issue #555, opened 2018-12-28
Support Kotlin/Native 


charleskorn:
I can see #327 added support for the multiplatform project model, but I can't see any references to Kotlin/Native.

Is this something you plan to support? If so, do you have an estimate for when?

And would you be open to a PR / PRs to add support for this?

raniejade:
It will be supported eventually. PRs are definitely welcome!

charleskorn:
Great to hear. Is there anything in particular you'd like to see a PR for? Based on a quick glance, it looks like the first step would be implementing the runtime and then some kind of test runner.

raniejade:
Yep, more specifically:
- how are tests discovered?
- how are tests executed?

Good place to start is the  `kotlin-test` implementation for K/N.  Is there a way to hook into it or something more native (think `mocha` and `jasmine` for JS).

charleskorn:
Last time I looked at `kotlin-test` in K/N, there was a compile-time operation that scanned for tests and effectively generated a `main()` that then ran them (there's a bit more to it than that, but that's the general idea). The reason for this is that K/N doesn't (and won't) support reflection, so some other mechanism is required. 

As an initial step, I'm thinking there could be something very simple that scans the source set for `Spek` classes and generates a `main()`, that can then be executed to execute and report on the tests. In the future this could be something more polished like a compiler plugin (for generating the `main()`) and a Gradle plugin (for hooking into Gradle's test infrastructure). How does that sound?

raniejade:
Ideally we'd want something similar for JVM (so we can drop `kotlin-reflect`) and JS.

>  I'm thinking there could be something very simple that scans the source set for Spek classes and generates a main()

I think the only option we have (if we don't start with the compiler plugin) is to parse the test souces which doesn't look very simple to me.

raniejade:
Did a quick spike on the compiler-plugin: https://github.com/raniejade/ct-discovery/blob/master/kotlin-plugin/src/main/kotlin/com/github/raniejade/ct/plugin.kt#L56

Basically I can see all classes being compiled and introspect annotations, super classes, etc. I just need a way to generate code.

charleskorn:
That's really cool, it doesn't look too difficult to create a compiler plugin then.

https://github.com/JetBrains/kotlin-native/blob/master/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/TestProcessor.kt seems to be where most of the magic happens for `kotlin-test`. 

[...]

PR #601, opened 2019-02-04
Initial support for Kotlin/Native

charleskorn:
See discussion in commit and #555 for more details.

Future PRs will need to add:
* better console output - #600 
* support for filtering tests (at the moment, all configured classes run)
* the Travis CI build needs to be configured to build on all three OSes and to then publish these binaries (Kotlin/Native doesn't support cross compilation) - #599
* a Kotlin compiler plugin to generate the `main()` automatically for test source sets, rather than the current hand-written one

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,6 @@ buildscript {
     }
 
     dependencies {
-        classpath 'io.spring.gradle:dependency-management-plugin:1.0.3.RELEASE'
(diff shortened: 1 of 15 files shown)


raniejade:
I've also filed #602, to refactor `Path` into the common module.

raniejade (approved):

PR #605, opened 2019-02-04
Correcting Test Snippets

alilosoft:
Replacing `it` with `test` in the discovery and fixtures examples.
Because `it` can't added directly to the `group` scope.

### DIFF
--- a/docs/core-concepts.md
+++ b/docs/core-concepts.md
@@ -37,14 +37,14 @@ object MyTest: Spek({
     println("this is the root")
     group("some group") {
         println("some group")
-        it("some test") {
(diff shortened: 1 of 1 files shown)


raniejade (approved):
Nice catch, thanks!

vanniktech (approved):

alilosoft:
> Nice catch, thanks!
Hi, I'm very glad that my PR get merged 🥇 , because actually I started playing with Spek only today and  started by copying snippets from the doc to my IntelliJ demo project to try them live, and those examples couldn't compile, so I followed the previous example to correct them and that works!

PR #607, opened 2019-02-04
Update to current year

ramonPires:
Just updating README.md year to 2019.


vanniktech (commented):
I believe this should not be done since it can invalidate the copyright

raniejade:
Can we just set it to `2018-2019`? Though, I have no clue about copyrights 🤷‍♂️

PR #609, opened 2019-02-05
Refactor DiscoveryContext API

### LINKED ISSUE #608: Rich DiscoveryContext API
The primary usage of this API is for discovering Test classes. Ideally it's an abstraction over platform specific Test discovery mechanisms. In JVM for example it could be backed by  [classgraph](https://github.com/classgraph/classgraph) (we already use it in the JVM implementation for `SpekRuntime`). On Native (and possible JS) due to the lack of runtime reflection it's backed by compile-time generated information (via a compiler plugin but initially hand rolled). With this API we can move `SpekRuntime` into the common module without the need of having platform specific modules.

```kotlin
interface TestInfo {
    val path: Path
    fun createInstance(): Spek
}

interface DiscoveryContext {
    fun getTests(): List<TestInfo>

    companion object {
        fun builder(): DiscoveryContextBuilder
    }
}

interface DiscoveryContextBuilder {
   fun addClass(cls: KClass<out Spek>, factory: () -> Spek): DiscoveryContextBuilder
   fun build(): DiscoveryContext
}
```

raniejade:
resolves #608.

At this point we don't need platform specific versions of `SpekRuntime`, I'll work on a separate PR to merge the two implementations.

//cc: @charleskorn

### DIFF
--- a/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/discovery/context.kt
+++ b/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/discovery/context.kt
@@ -6,23 +6,35 @@ import org.spekframework.spek2.runtime.scope.PathBuilder
 import kotlin.reflect.KClass
 
 
-typealias SpekProducer = () -> Spek
+class TestInfo(val path: Path, private val factory: () -> Spek) {
+    fun createInstance(): Spek {
+        return factory()
+    }
+}
 
-data class DiscoveryContext(val classes: Map<KClass<out Spek>, SpekProducer>) {
+class DiscoveryContext(private val testInfos: List<TestInfo>) {
+    fun getTests(): List<TestInfo> {
+        return testInfos
+    }
+
     companion object {
         fun builder(): DiscoveryContextBuilder = DiscoveryContextBuilder()
     }
 }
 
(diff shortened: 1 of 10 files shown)


spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/discovery/context.kt
```diff
@@ -6,23 +6,35 @@ import org.spekframework.spek2.runtime.scope.PathBuilder
...
+    fun addTest(cls: KClass<out Spek>, factory: () -> Spek): DiscoveryContextBuilder {
+        val path = PathBuilder.from(cls)
+            .build()
+        classInfos.add(TestInfo(path, factory))
         return this
     }
 
-    fun build(): DiscoveryContext = DiscoveryContext(classes)
-}
+    inline fun<reified T: Spek> addTest(noinline factory: () -> T): DiscoveryContextBuilder {
```
charleskorn, line 35: Very minor formatting thing: there's no space between `fun` and the `<reified...`

(might be worth setting up something like Spotless to check for these kinds of things)

charleskorn (approved):

charleskorn:
Looks good to me. Merging the runtimes sounds like a great idea.

Issue #610, opened 2019-02-05
Merge SpekRuntime implementations


raniejade:
With #608, platform specific discovery has been abstracted making it possible to have a single implementation of `SpekRuntime`.

PR #617, opened 2019-02-06
Fix instrinsic NPE in Java 1.8

### LINKED ISSUE #574: Couldn't run test clicking in the greens arrows alongside of the class
This morning I found an update from Spek (to **v2.0.0-rc.1.180+b8533a4-IJ2018.3**), and after did it I couldn't run any test individually (**clicking in the green arrows alongside the class**). The only way that I have found to run the test is running **all** the test classes that I have in the folder **at the same time** (right-click on the folder -->Run Test in whatever).
 
I'm working with Windows and IntelliJ IDEA 2018.3.3 (Ultimate Edition, Build #IU-183.5153.38) 
This bug was easy to reproduce as well. 

Please, find the images below for more details:

annotation 2019-01-23 114226

This is the trace:

annotation 2019-01-23 114259

raniejade:
Resolves #574 

Only happens to classes in the default package.


vanniktech (approved):
Can we somehow reuse the code between the runner and the runtime?

raniejade:
I'll do it as part of #610.

PR #623, opened 2019-02-08
Create separate paths for every package part

### LINKED ISSUE #606: Error "No test were found" when running tests from a package that contains other packages
Environment:
Android Studio 3.3Build #AI-182.5107.16.33.5199772, built on December 25, 2018
Spek plugin versoin: 2.0.0-Studio3.3
spekVersion = "2.0.0"

Prerequirements:
a package that contains other packages with unit tests.

Steps to reproduce

1. Right click on a package-> Run 'Tests in '[package_name]''
2. "Instantiating tests... " is displayed in the Run tab
3.  "No test were found" is displayed in the Run tab

Also, the first step works only from the second attempt, so after the first attempt nothing happens - test suite is not launched.
Generally, the bug behavior is similar to https://github.com/spekframework/spek/issues/543

raniejade:
Resolves #606.

Basically changes the path generation for package names.

`com.example/Foo` -> `com/example/Foo`


vanniktech (approved):

raniejade:
@vanniktech so this actual broke compatibility between `2.0.1` and `2.0.0`. The old version is sending an old path format. I'm honestly lost on how to fix it, most of the solution I came up are hacks that may introduce other bugs. Given that #606 is a major bug, I'm happy to break compatibility here. WDYT?

PR #625, opened 2019-02-08
Speed up builds

raniejade:

PR #626, opened 2019-02-08
Speed up build without travis changes

raniejade:
- Gradle 5.2
- `org.gradle.parallel=true`
- Kotlin 1.3.21

Building with travis changes actually more slower because of: https://github.com/JetBrains/gradle-intellij-plugin/issues/123


vanniktech (approved):
Is parallel safe on CI?

raniejade:
from what I can see yes, I'll experiment more with #625 and see how I can speed it up more. 20mins to build sucks :/

PR #635, opened 2019-02-14
Correct Spek version interpolation for Gradle

seadowg:
Gradle setup instructions contained incorrect syntax


raniejade (approved):
Thanks!

Issue #639, opened 2019-02-23
Report better test case names on failures or provide a mechanism to fetch the full path name of the test


dibog:
If a test case fails the reports are not very descriptive.

E.g. take: 

```
class CalculatorSpec : Spek({
    describe("Calculator") {
        val calculator by memoized { Calculator()}

        listOf(
                TestArg( 1, 2, 3+1 ), // <- to provoke the test cases to fail
                TestArg( 2, 3, 5+1 ),
                TestArg( 4, 5, 9+1 )
        ).forEach { ( a, b, expectedSum ) ->
            context("adding $a and $b") {
                val actualSum = calculator.add(a, b)

                it("should be $expectedSum") {
                    assertThat( actualSum ).isEqualTo( expectedSum )
                }
            }
        }
    }
})
```
which will report those failures as
```
[ERROR]   CalculatorSpec expected:<[10]> but was:<[9]>
[ERROR]   CalculatorSpec expected:<[4]> but was:<[3]>
[ERROR]   CalculatorSpec expected:<[6]> but was:<[5]>
```

The message would be better if we can provide the path name of the test.
Unfortunately this path name can't be accessed currently. With the following,
very ugly extension function:
```
fun TestBody.fullPathName(): String? {
    val testScope = memoized<TestScope>() as? MemoizedValueReader<TestScope>
    return testScope?.scope?.path.toString()
}
```
and the assertion in the original code changed as the following:
```
assertThat( actualSum, fullPathName() ).isEqualTo( expectedSum )
```
we get a better report:
```
[ERROR]   CalculatorSpec expected [test/CalculatorSpec/Calculator/adding 4 and 5/should be 10]:<[10]> but was:<[9]>
[ERROR]   CalculatorSpec expected [test/CalculatorSpec/Calculator/adding 1 and 2/should be 4]:<[4]> but was:<[3]>
[ERROR]   CalculatorSpec expected [test/CalculatorSpec/Calculator/adding 2 and 3/should be 6]:<[6]> but was:<[5]>
```

This is IMHO more readable.

So please either let the report generate a better description or add the fullPathName() method to the TestBody.

raniejade:
IMHO generating more descriptive reports is the way to go.  Starting `2.1.0` as part of Kotlin Native support - Spek will provide its own gradle plugin which will have proper reporting (similar to rspec or jasmine).

dibog:
[...]

PR #641, opened 2019-02-24
adding TestBody.testName() to get a human readable representation of …

dibog:
…the test to be used within assertions

Please find here my pull request. If you point me where to add unit cases for this I will happy to add them to the pull request.

See issue: https://github.com/spekframework/spek/issues/639


raniejade:
@dibog thanks for the PR, but I don't think this is the right way to go. Adding `testName` is just a stop gap, it doesn't add any value other than your use case.

dibog:
It helps if you want to use the name of the test case in your assertions. Most assertion framework supports this.

PR #647, opened 2019-03-06
Implement test timeouts

### LINKED ISSUE #643: Test timeouts
Spek executes tests serially in a single thread, any rogue tests that blocks indefinitely can cause the whole execution to block. Introducing an execution timeout per test will help stop that from happening. 
- timeouts should be configurable globally and a per test basis.
- timeouts should only be applied to logic that participate in the execution phase: fixtures and test scopes.

Implementation details:
- `beforeEachTest`, `afterEachTest` and the `test` itself is considered as a single unit, so the timeout is applied to their combined execution time.
- `beforeGroup` and `afterGroup` timeout semantics is undefined for now.

raniejade:
Resolves #643 

An optional timeout parameter is added to `test` scope builder which defaults to 10 seconds. Timeout is implemented using coroutines under the hood, this means that timeout is not working on KN (only a single threaded coroutine is supported). Also, note that timeouts do not apply to `beforeGroup`, `afterGroup` and `afterEachTest`.


spek-runtime/build.gradle
```diff
@@ -49,14 +51,23 @@ kotlin {
 
         linuxMain {
             dependsOn nativeMain
+            dependencies {
+                implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core-native:1.1.1'
```
raniejade, line 55: `dependencies.gradle` does not work here for some reason.

integration-test/src/jvmTest/kotlin/org/spekframework/spek2/integration/TimeoutTest.kt
```diff
@@ -0,0 +1,16 @@
+package org.spekframework.spek2.integration
+
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.meta.Ignore
+
+// remove @Ignore to test if timeouts are working.
+@Ignore
```
raniejade, line 7: I don't want the build to fail, just trust me that it's working :)

settings.gradle
```diff
@@ -15,6 +15,7 @@ pluginManagement {
 }
 
 enableFeaturePreview('STABLE_PUBLISHING')
+enableFeaturePreview('GRADLE_METADATA')
```
raniejade, line 18: required by the coroutines library.

boylenssen:
While debugging, I'm not able to take my time as this timeout occurs. I'm using the Gherkin method, so where can I override the timeout? And isn't there a way to set the timeout globally?

raniejade:
@boylenssen just filed https://github.com/spekframework/spek/issues/692 and will be included in `2.0.5`

spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/Collectors.kt
```diff
@@ -71,11 +73,12 @@ class Collector(
 
     }
 
-    override fun test(description: String, skip: Skip, body: TestBody.() -> Unit) {
+    override fun test(description: String, skip: Skip, timeout: Long, body: TestBody.() -> Unit) {
         val test = TestScopeImpl(
             idFor(description),
             root.path.resolve(description),
             root,
+            defaultTimeout,
```
fdc-liebreich, line 81: you want to use the argument here instead

raniejade: yep, already fixed in master.

boylenssen: so I have this: 
```
class ResponseTest : Spek({

    setArchTaskExecutor()

    timeout(15000)
```
but that doesn't work. How to I set it?

raniejade: `defaultTimeout = xxx`

Issue #658, opened 2019-03-20
Eagerly evaluated alternative to memoized.


feature:
The syntax of `memoized` values provide a short and concise way of storing values, and it is much nicer than using lateinit vars and a `beforeEach` blocks. However because it is lazily evaluated it can cause some strange issues, as shown in my previous issue (#645) and down below.

Today I encountered into another issues with a combination of `mockito` and spek's `memoized`. If you memoize a mock object `A` that stubs a function, and you memoize another mock `B` that stubs a function that returns mock `A`, mockito will fail because the memoized values are lazily evaluated and therefore mock `A` will be stubbed while `B` is being stubbed, which is not allowed by mockito. My solution to this problem was the same as with my previous issue; use lateinit vars and a `beforeEach` block.

The issue I just describe isn't very related to spek, however I still think spek should consider adding an alternative to `memoized` to make these things easier and cleaner than what we current have to do.

Related PR: #660

raniejade:
Can you add a small snippet of code? I think it will make things clearer to understand.

feature:
```kotlin
val a by memoized {
    mock<Any>() {
        // stub something
    }
}
val b by memoized {
    mock<Any>() {
        // stub something that uses `a`
    }
}
```

Because `a` and `b` are both memoized they aren't initialized until they are used. As `b` depends on `a` when its mock is created and being stubbed, `a` will also be created and stubbed while `b` is being stubbed. This behavior is disallowed by Mockito for some reason, and will throw an exception.

My issue isn't meant to be focused on this issue, but rather the suggestion of adding a new construct to spek for simplifying a common pattern, a pattern spek has already deemed common enough for adding the `memoized` function, thought I think the implementation of `memoized` hinders its usefulness.

raniejade:
One option is to add a `lazy` parameter to `memoized`, which defaults to `true`. I'll take a look at it more when I get back from vacation next week. Feel free to pull up a PR that explores that or any idea you might have.

Issue #645, opened 2019-03-03
memoized produces unexpected results


feature:
The documentation for [scope values](https://spekframework.org/core-concepts/#scope-values) gives the impression that the `memoized` delegate is just a shorthand for `lateinit var`s and `beforeEach` blocks. However if the function passed to the `memoized` modifies another memoized value, the change is not reflected in the value.

I'm not sure if this is a bug with `memoized` or not, because I can't find any documentation on the expected behavior of the function. Regardless this is unintuitive, and that behavior is not documented.

Particularly I'm testing [this](https://github.com/staminadevelopment/pubsub4k/blob/master/src/main/kotlin/pw/stamina/pubsub4k/publish/StandardPublisherRegistry.kt#L40) function from a library I'm working on.

In my tests the follow code works as expected:
```kotlin
lateinit var publisher: Publisher<Any>
beforeEach {
    publisher = registry.findOrCreatePublisher(topic) { subscriptions }
}
```
But using `memoized` does not:
```kotlin
val publisher by memoized { registry.findOrCreatePublisher(topic) { subscriptions } }
```

Read comment: https://github.com/spekframework/spek/issues/645#issuecomment-469019149

raniejade:
Can you paste in the full test?

feature:
I was still working on the tests, but here they are. All the affected cases are commented with `Do not use memoized here`.

https://github.com/staminadevelopment/pubsub4k/blob/master/src/test/kotlin/pw/stamina/pubsub4k/publish/StandardPublisherRegistrySpec.kt

raniejade:
> However if the function passed to the memoized modifies another memoized value, the change is not reflected in the value.

Did you mean _access_? You can't modify a memoized value.

Anyway, the exact semantics are (and you're right I need to add more documentation around it)

```
    val a by memoized { 1 }
    val b by memoized { a + 1 }
```

If `b` was accessed first then the value is initialized by calling the passed factory lambda, since it's accessing another memoized value `a`,  then the same thing will happen with `a`.

This can be seen by doing some simple `println`s.

```kotlin
    val a by memoized {
        println("a")
        1
    }
    val b by memoized {
        println("b")
        return@memoized { a + 1 }
    }

    test("foo") {
        println(b())
    }
```

The output will be:
```
b
a
2
```

`memoized` should work with your use case, so I'm probably missing something here.

feature:
I'm not trying to change the value of a memoized property, but make changes to the object that is memoized.

[...]

PR #660, opened 2019-03-21
Add eagerly evaluated memoized

feature:
This PR adds a new type of `memoized` values that are eagerly evaluated. The goal of this PR is the fix the issues presented in #658 and #645. Tests still needs to be added, and I'd greatly appreciate some help with help how to properly test this feature.

The code works so far that values are eagerly evaluated for each test, however I discovered that the order in which they are evaluated is reversed. I found this is because the `LifecycleListener` for each memoized value being [*prepended* to the list of listeners](https://github.com/spekframework/spek/blob/2.x/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/lifecycle/LifecycleManager.kt#L17). I'm hesitant to changing this to appending the listeners, as I'm not sure if it will break existing code relying on Spek. Additionally all fixtures are grouped together in a single listener, that's always invoked last. This means you can't mix fixtures before or between eager memoized.

The following code should help illustrate the ordering issue. A user might expect the printed result to be `A2C1B3`, but because of the implementation details describe above the actual result is `BCA213`.
```kotlin
val a by eagerMemoized { print("A") }
beforeEach {             print("2") }
val c by eagerMemoized { print("C") }
beforeEach {             print("1") }
val b by eagerMemoized { print("B") }
beforeEach {             print("3") }
```

### DIFF
--- a/spek-dsl/src/commonMain/kotlin/org/spekframework/spek2/dsl/dsl.kt
+++ b/spek-dsl/src/commonMain/kotlin/org/spekframework/spek2/dsl/dsl.kt
@@ -29,6 +29,9 @@ interface LifecycleAware : ScopeBody {
     fun <T> memoized(mode: CachingMode = defaultCachingMode, factory: () -> T): MemoizedValue<T>
     fun <T> memoized(mode: CachingMode = defaultCachingMode, factory: () -> T, destructor: (T) -> Unit): MemoizedValue<T>
 
+    fun <T> eagerMemoized(mode: CachingMode = defaultCachingMode, factory: () -> T): MemoizedValue<T>
+    fun <T> eagerMemoized(mode: CachingMode = defaultCachingMode, factory: () -> T, destructor: (T) -> Unit): MemoizedValue<T>
+
     fun beforeEachTest(callback: () -> Unit)
     fun afterEachTest(callback: () -> Unit)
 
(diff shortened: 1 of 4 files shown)


feature:
@raniejade Do you think it would be possible to add the listeners for memoized values as fixtures, instead of having own `LifecycleListener`s? That should solve all the issues with ordering, and I don't think it would introduce any issues for projects already using Spek. Please give me your thoughts on this.

An alternative, and probably even better idea is to add the initialization of eager memoized values as fixtures. That would fix the ordering issues, without changing any existing behavior. This would also as closely as possible emulate using lateinit vars and fixtures.

raniejade:
Food for thought: just make `memoized` eager, and evaluate based on declaration order. wdyt? I've opened up a poll in the #spek slack channel.

feature:
In my option that seems to be a good solution, but I figured you probably wouldn't do that due to potential backwards compatibility issues. As of of 5913e65006f329678db3cb3fe51c2fec7de7e61f I've implemented this feature, and it seems to work without any issues, so now the question is just if we want to update `memoized`, or create a new construct like `eagerMemoized`.

raniejade:
I'm leaning more into making `memoized` eager as it is simpler, having two ways to use scope values just complicates things. I don't think removing laziness is a big breaking change, the values will be evaluated eventually. 

Making it eager means, this:

```kotlin
val foo by memoized { Foo() }
val bar by memoized { Bar() }

it("do something") { 
  assertEquals(foo.something(), bar.something())
}
```
is _roughly equivalent_, in terms of execution, to:

```kotlin
val foo = Foo()
val bar = Bar()
assertEquals(foo.something(), bar.something())
```

If you don't mind, can you create a separate PR that makes `memoized` eager?

feature:
Sure thing champ.

PR #661, opened 2019-03-25
Make memoized eagerly evaluated

feature:
This PR makes `memoized` values be eagerly evaluated. For more information this please check out PR #660.

### DIFF
--- a/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/lifecycle/MemoizedValueAdapter.kt
+++ b/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/lifecycle/MemoizedValueAdapter.kt
@@ -1,5 +1,6 @@
 package org.spekframework.spek2.runtime.lifecycle
 
+import org.spekframework.spek2.dsl.Root
 import org.spekframework.spek2.lifecycle.GroupScope
 import org.spekframework.spek2.lifecycle.LifecycleListener
 import org.spekframework.spek2.lifecycle.TestScope
@@ -9,8 +10,7 @@ import kotlin.reflect.KProperty
 
 sealed class MemoizedValueAdapter<T>(
     val factory: () -> T,
-    val destructor: (T) -> Unit,
-    val eager: Boolean
+    val destructor: (T) -> Unit
 ) : ReadOnlyProperty<Any?, T>, LifecycleListener {
 
     protected sealed class Cached<out T> {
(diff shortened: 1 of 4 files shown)


spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/lifecycle/MemoizedValueAdapter.kt
```diff
@@ -33,51 +34,52 @@ sealed class MemoizedValueAdapter<T>(
...
-        override fun afterExecuteGroup(group: GroupScope) {
-            val cached = this.cached
-            if (cached is Cached.Value<T>) {
-                destructor(cached.value)
-            }
-            if (stack.isNotEmpty()) {
-                this.cached = stack.removeAt(0)
-            }
+    override fun registerEagerInitializer(root: Root) {
+        root.beforeGroup {
```
raniejade, line 48: Is this due to the runtime pre-pending lifecycle listeners? https://github.com/spekframework/spek/blob/2.x/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/lifecycle/LifecycleManager.kt#L17

feature: Yes. As I explained in the previous PR, there were two ordering issues, this seems to be the easiest way to solve both of them, to keep the order of `memoized` initialization and fixtures consistent with their declaration order in tests.

feature:
I'm able to fully build the entire project on my machine. It seems Travis CI is just timing out due to time constraints, and therefore the build fails for my commits. Please confirm for yourselves first before merging.

raniejade:
@feature sorry for not being responsive, I'll get to this once I get some free time (hopefully by this weekend or next week).

spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/lifecycle/MemoizedValueAdapter.kt
raniejade: I'll get back to you on why I did it(prepend) that way (I totally forgot). I'm thinking if we could just change `memoized` to use the fixtures or change the implementation of fixtures so that you can weave them in between `memoized` calls and get the _expected_ execution sequence.

raniejade: found it https://github.com/spekframework/spek/pull/183/files 🤦‍♂️ , what was I thinking.. Anyway we have to ways to fix this:

1. Change `memoized` implementation to use fixtures.
2. Change fixtures implementation to register a LifecycleListener for each call instead of having one LifecycleListener (`FixturesAdapater`)- and stop prepending listeners.

If we can make sure that the execution order of `memoized` and `fixtures` are based on declartion order then we can stop prepending listeners. I'm leaning more to option 1, wdyt?

feature: I apologize for the delayed response. Both options works fine to me, though implementing `memoized` using fixtures seems the easiest, and is the least likely to break backwards compatibility. I need to mention that memoized still registers its own `LifecycleListener`s for clearing its values and invoking the destructors. I think we should either use fixtures for invoking the destructors, or keep them as listeners, and do as you described in `2.`.

feature: @raniejade I'm sorry to tag you, but you haven't responded in 4 days now.

raniejade: @feature sorry for the delayed response! I'm leaning towards using fixtures all the way, so instead of `memoized` registering `LifecycleListerner`s it just register fixtures (depending on the `CachingMode`).

raniejade: @feature just realized that we cant use fixtures with `CachingMode.GROUP`. `beforeGroup` is only executed once not per group.

feature: Hmm. Do you got any other ideas? I have been rather busy with life, but I'll take another deeper look in the issue when I have time.

raniejade: Trying to align fixtures and memoized, eventually adding a `beforeEachGroup` fixture. I think that should help us implement this.

jcornaz:
What's the status of this?

raniejade:
@jcornaz Making memoized eager is definitely going to happen, I'm still thinking on the best approach to do this.

raniejade:
`2.x` branch has diverged much since this PR was raised, closing it.

PR #741, opened 2019-07-01
Publish JS dsl artifacts

### LINKED ISSUE #706: Provide JS and Native api artifacts
This will allow MPP projects to use spek smoothly even though tests aren't run yet.

raniejade:
Resolves #706

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -82,7 +82,7 @@ stages:
     dependsOn: test_ide_plugins
     condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'), eq(variables['Build.SourceBranch'], 'refs/heads/2.x'))
     jobs:
-      - job: jvm_and_linux
+      - job: jvm_js_and_linux
         pool:
           vmImage: 'ubuntu-16.04'
         steps:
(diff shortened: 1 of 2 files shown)


robstoll:
@raniejade I tried it out for KBox, worked as expected. However, for Atrium it does not. It fails with `Its dependency 'kotlin' was not found. ` even though other modules are loaded before hand and include the same check ` if (typeof kotlin === 'undefined') { ...`. I guess the problem comes from different JS module systems but that's only a guess. Just something you can keep in mind in case you start implementing a runner for JS (I could contribute as well in case you can need help)

raniejade:
> I guess the problem comes from different JS module systems but that's only a guess.

Can you explain more? I'll see if I can think of something.

robstoll:
I am referring to this https://kotlinlang.org/docs/reference/js-modules.html
The tests in Atrium require to use commonjs in order that mocha can pick up the tests

raniejade:
Hmmm, weird. The JS artifact is published as an UMD module so it should be compatible with common js.

robstoll:
yeah... it should be, I remember that I ran into similar problems already. Maybe a Kotlin bug, I'll file an issue and see what they are saying (https://youtrack.jetbrains.com/issue/KT-33611)

PR #745, opened 2019-07-07
Fix line marker for non-direct spek subclass

### LINKED ISSUE #721: [IDEA] abstract class without explicit constructor, subclasses cannot be run
Following an example:
```
abstract class C : Spek({
    describe("this one") {
        it("can be run") {
            //so that we see it is actually also run in gradle
            throw IllegalStateException("hello")
        }
    }
})

//gutter icon is missing for both of them -> change `C` above  to `C()` and the gutter icon appears
class D : C()
object E: C() 
```

raniejade:
Resolves #721


spek-ide-plugin-intellij-base/src/main/kotlin/org/spekframework/intellij/domain/ScopeDescriptorCache.kt
```diff
@@ -31,7 +31,7 @@ class ScopeDescriptorCache: ProjectComponent {
     }
 
     fun fromClassOrObject(clz: KtClassOrObject): ScopeDescriptor.Group? {
-        if (!isSpekSubclass(clz) || clz.isAbstract()) {
+        if (clz.isAbstract() || !isSpekSubclass(clz)) {
```
raniejade: Avoid unnecessary hierarchy checks when class is abstract.

PR #753, opened 2019-07-18
Fail fast gherkin style

### LINKED ISSUE #714: Fail fast with Gherkin style
Hi, 

When we use Gherkin style to write a scenario, if a `Given` or a `When` fail (throw an exception), there is no much point to run the assertions in the `Then`.

Actually, running them even make more difficult to interpret the result of the test IMHO. Because we may have many assertions (in many `Then` and `And`), we'll see many assertion failures, and may not notice at a glance that the `When` where actually throwing an exception.

jcornaz:
resolves #714 

Written during [Basel hackergarten](http://hackergarten.net) with @anneloreegger, @dpdawson and @ursjoss


raniejade (approved):
🎉 thanks guys! I'll include it the `2.0.7` release.

PR #766, opened 2019-08-16
add `include` to Root

robstoll:
I have used `include`now in various cases in [Atrium](https://atriumlib.org) and it seems stable to me, thus pushing it back here.


raniejade (approved):
Thanks!

PR #772, opened 2019-08-25
Update setup-android.md

erikhuizinga:
Move `unitTests.all` closure to the `testOptions` closure on which it operates.

It seemed that it operated on the `jacocoOptions` closure, but it doesn't.


raniejade (approved):
ooh, thanks!

PR #779, opened 2019-09-01
Add 'Share common logic between tests' doc section

### LINKED ISSUE #777: Inheritance hierarchy: Base class support in Spek
I have the following use case.
I implemented an abstract base class, which creates expensive objects before each test and disposes them after each test. To have this boilerplate code not spread through the code base, I created the following base class. Other `Spek` classes derive from this base class called `BaseSpek`.

How can I access the `object` variable in derived classes in order to call methods on this object? Is there maybe another way to accomplish my objective? Is this use case even supported by the Spek framework?

```Kotlin
abstract class BaseSpek(root: Root.() -> Unit) : Spek({

    // how can I access this variable in a class that derives from 'BaseSpek'
    var object: ExpensiveObject? = null
    
    beforeEachTest {
        object = createExpensiveObject()
    }
    
    afterEachTest {
        object.dispose() // dispose manually
    }
    
    root()
})
```

schalkms:
Closes #777

### DIFF
--- a/docs/core-concepts.md
+++ b/docs/core-concepts.md
@@ -153,3 +153,28 @@ You can pass in an optional parameter to `memoized` which controls how the value
 - `CachingMode.INHERIT`: internal use only.
 
 If you are using the [gherkin](gherkin.md) style note that the default caching mode is `CachingMode.GROUP`.
+
+### Share common logic between tests
+Often there is some code, which needs to be called repeatedly before and after tests.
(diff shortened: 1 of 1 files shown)


schalkms (commented):
What's wrong with the build?

raniejade:
Not sure, but I just restarted the checks.

raniejade:
Thanks! :)

PR #785, opened 2019-09-10
Add execution result to LifecycleListener after scope methods

### LINKED ISSUE #761: Add custom ExecutionListener/Reporter
Is there anyway to add a custom reporter or simply to listen to an event "group failed".

In the `LifecycleListener`, we don't get any information in `afterExecuteGroup` about the result 🤔 

Thanks.

raniejade:
Resolves #761.

`LifecycleListener#afterExecute{Test|Group}` now includes an `ExecutionResult` (Success | Failure(Throwable)).

Some implementation details:
- `afterEachTest` and `afterGroup` fixtures are invoked last (just after lifecycle listeners are invoked), if any of the fixtures throws an exception the reported ExecutionResult may be a `Success`.
- `afterExecuteGroup` will only report failures when an exception is thrown in a `beforeGroup` fixture. It will not aggregate the result of its children.

### DIFF
--- a/integration-test/build.gradle.kts
+++ b/integration-test/build.gradle.kts
@@ -34,6 +34,8 @@ kotlin {
 
             compilations["test"].defaultSourceSet {
                 dependencies {
+                    implementation(Dependencies.mockitoKotlin)
+                    implementation(Dependencies.mockitoCore)
                     runtimeOnly(kotlin("reflect"))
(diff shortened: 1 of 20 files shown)


raniejade:
@badoualy can you take a look? :pray:

badoualy (approved):
Looks good, shame that we end up with two `ExecutionResult`. Should probably be refactored when bumping to a new major version.

Thanks!

PR #819, opened 2019-10-25
Added a rule to fail scopes with empty descriptions

### LINKED ISSUE #689: Disallow empty descriptions
When running via `gradle test` no spek tests will be run when a block (group, describe, it etc) has an empty description. 

This issue does not happen when running via the spek framework plugin.

Spek version: 2.0.1 - 2.0.4 (possibly any 2.x version)
Android studio: 3.4

marvinbuff:
Resolves #689

I implemented this feature as a follow up to the Basel HackerGarten. 
Hope it is useful and correct. If you have any suggestions or requests, I will try to comply.

### DIFF
--- a/integration-test/src/jvmTest/kotlin/org/spekframework/spek2/FailureTest.kt
+++ b/integration-test/src/jvmTest/kotlin/org/spekframework/spek2/FailureTest.kt
@@ -13,6 +13,13 @@ object FailureTest: AbstractSpekTest({ helper ->
                         test("4 == 2 + 2")
                         test("false == true", false)
                     }
+                    group("empty description") {
+                        test("", false)
+                        test("", false)
+                    }
+                    group("") {
+                        test("4 == 2 + 2", false)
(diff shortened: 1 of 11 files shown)


raniejade (changes requested):
Thanks for the PR! I think the checks should be done in [Collector](https://github.com/spekframework/spek/blob/2.x/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/Collectors.kt).

marvinbuff:
> Thanks for the PR! I think the checks should be done in [Collector](https://github.com/spekframework/spek/blob/2.x/spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/Collectors.kt).

Done. The fail behaviour is now a little bit different. I used the structure in-place but maybe you want it differently. Just let me know.

raniejade (approved):
Looks good, thanks for the contribution!

Issue #789, opened 2019-09-11
Make scope value access stricter


raniejade:
The current implementation allows scope value with a `CachingMode.TEST` to be accessed inside a `afterGroup` fixture. Consider the following scenario:

```kotlin
val a by memoized { FooService() }

test("some test") {
  a.doSomething()
}

afterGroup {
	a.cleanup()
}
```

Scope values with `CachingMode.TEST` are cleared after each test execution, so the value of `a` in `afterGroup` is not the same as the value when `doSomething` was called. In this case, cleanup was actually called on a new instance!

It gets a bit weird when accessing the same scope value in a `beforeGroup`.

```kotlin
val a by memoized { FooService() }

beforeGroup {
	a.setup()
}

// assume that a.doSomething() depends on a.setup() being called first, otherwise it will throw an exception

test("some test") {
  a.doSomething()
}

test("some test 2") {
  a.doSomething()
}
```

In this case the test `some test 2` will fail!

The second scenario is noticeable and can be addressed because the test will fail, while the first scenario can be completely missed!

To avoid completely miss addressing the first scenario we need to make sure scope values will throw an exception when accessed in the wrong context instead of just blindly creating a new value.

PR #835, opened 2019-12-10
Sprinkle the suspend keyword wherever it makes sense.

### LINKED ISSUE #426: Add support for tests with coroutines
Would it be possible to add `cdescribe`, `ccontext`, `cBeforeEachTest` and `cit`, etc.. (for example), that would have a `runBlocking` (or the new TestCoroutine system...) inside... that would avoid extra nesting (like mockk.io did...).

raniejade:
WIP: most likely contains bugs.

### The following blocks are now _suspending_:
- tests (`its`, `Then`s, etc...)
- fixtures (`beforeEachTest`, etc...)
- scope value's constructor and destructor

### Scope values are eagerly invoked now.
This won't be an issue if your tests are passing on Spek `2.0.8`. Since that version (#789) accessing scope values became stricter which is in preparation for this. This change is necessary to be able to make the scope value's constructor and destructor suspending.

---

Big bonus with this change: integration tests regarding the timeout feature runs more faster now :)

Resolves #426


jyodroid:
sorry probably is the wrong place but i'm totally lost, how can use spek to test coroutines? any example on repository?

raniejade:
Hey @jyodroid, this hasn't been released yet - scheduled for `2.1.0`.

PR #840, opened 2020-01-06
Add support for RSpec `let`-style contextually overridable values.

### LINKED ISSUE #807: Support let, subject methods like in RSpec
I know Kotlin has different constraints, but subject and let are very valuable to remove boilerplate code. Just wanted to reference this as a potential feature.

See: 
https://relishapp.com/rspec/rspec-core/v/3-9/docs/subject
https://relishapp.com/rspec/rspec-core/v/3-9/docs/helper-methods/let-and-let

xian:
Use
```kt
val varName by value { expression }
```
to declare a _let value_, and
```kt
value(varName) { expression }
```
to override its value within a context. Values are memoized per-test and accessed by invocation:
```kt
println(varName())
```

Fixes #807.

### DIFF
--- a/integration-test/src/commonTest/kotlin/org/spekframework/spek2/integration/LetValueMetaSpec.kt
+++ b/integration-test/src/commonTest/kotlin/org/spekframework/spek2/integration/LetValueMetaSpec.kt
@@ -34,7 +34,7 @@ object LetValueMetaSpec : AbstractSpekTest({ helper ->
 
             val (_, result) = results.afterExecuteGroupResults.single()
             result as ExecutionResult.Failure
-            expect("letValue() can't be used from beforeEachGroup or afterEachGroup") { result.cause.message }
+            expect("letValue can't be used from beforeEachGroup or afterEachGroup") { result.cause.message }
         }
 
         it("should fail silently when invoked in afterEachGroup, cuz spek") {
@@ -45,7 +45,7 @@ object LetValueMetaSpec : AbstractSpekTest({ helper ->
             result as ExecutionResult.Success
             expect(listOf(
                     "afterEachGroup called",
-                    "letValue() failed: letValue() can't be used from beforeEachGroup or afterEachGroup"
+                    "letValue() failed: letValue can't be used from beforeEachGroup or afterEachGroup"
             )) { events }
         }
     }
@@ -54,7 +54,7 @@ object LetValueMetaSpec : AbstractSpekTest({ helper ->
 class LetValueInvokedFromBeforeEachGroupExample(private val listener: LifecycleListener) : Spek({
     registerListener(listener)
     val letValue by value { "anything" }
-    beforeEachGroup { letValue() }
+    beforeEachGroup { letValue.capitalize() }
     test("empty test") {}
 })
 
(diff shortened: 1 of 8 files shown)


xian:
Hello Spek maintainers! Here's a quick stab at implementing contextually overridable variables, let me know what you think. First PR here so would love any guidance on style or substance. Thanks!

xian:
@raniejade any thoughts here? Thanks!

raniejade:
Hey @xian, sorry I haven't looked at it yet. Maybe sometime next week after I'm done with an exam. At first glance it looks very similar to `memoized`, what is the problem it is trying to fix?

xian:
No problem. :-)

It's similar to `memoized` but it allows for nested contexts to override the value, which is then available in outer `before`/`after` blocks:

```kt
describe("page behavior") {
  val isAdmin by value { false }
  beforeEachTest { logIn(user, isAdmin()) }

  it("shouldn't include admin tab") {
    expect(false) { pageTabs.contains("Admin") }
  }

  context("when user is an admin") {
    value(isAdmin) { true }

    it("should include admin tab") {
      expect(true) { pageTabs.contains("Admin") }
    }
  }
}
```

Some more examples [here](http://www.betterspecs.org/#let).

raniejade:
Hey! I'm not convince about the value it adds. Your example can be re-written as:

```kotlin
describe("page behavior") {
  context("when user is not an admin") {
    beforeEachTest { logIn(user, false) }
    it("shouldn't include admin tab") {
      expect(false) { pageTabs.contains("Admin") }
    }
  }

  context("when user is an admin") {
    beforeEachTest { logIn(user, true) }
    it("should include admin tab") {
      expect(true) { pageTabs.contains("Admin") }
    }
  }
}
```

which is more readable in my opinion. There is a boilerplate with regards to doing the actual logging in, however, I think it helps with readability. To remove the boilerplate in the above example you can use a `var` and `beforeGroup` together:

```kotlin
describe("page behavior") {
  var admin = false
  beforeEachTest { logIn(user, false) }

  context("when user is not an admin") {
    it("shouldn't include admin tab") {
      expect(false) { pageTabs.contains("Admin") }
    }
  }

  context("when user is an admin") {
    beforeGroup { admin = true }
    it("should include admin tab") {
      expect(true) { pageTabs.contains("Admin") }
    }
  }
}
```

xian:
Fair enough, it's a subjective/stylistic thing. The `beforeGroup` approach looks clunky to me, and my experience with rspec's `let` is that it allows for superbly readable tests, but I'm not going to convince you of how much awesomer it is in github comments. (Fwiw we wanted to put it in Jasmine but JS gives no clean way to do it. Kotlin is so awesome.)

Anyway, would be cool if it could be implemented locally; would you be open to a minimally invasive PR that would allow for that?

raniejade:
Maybe before proceeding, shoot up an GH issue first and describe what you want to do. Let's also move the discussion on that. Or maybe just hijack #807.

O330oei (commented):
cont..

raniejade:
Closing this for now - I'm still not convinced with this feature - `let` is already implemented in the form of `memoized`. If you can come up with more use case please post it on #807.

PR #866, opened 2020-03-08
Fix kotlin native tests

raniejade:
//cc @charleskorn


integration-test/src/nativeMain/kotlin/Dummy.kt
```diff
@@ -1,2 +1,2 @@
 // This is a workaround for https://github.com/JetBrains/kotlin-native/issues/2617
-object Dummy
+object Dummy
```
raniejade, line 2: issue close now, but it didn't make it on the `1.3.70` release. So either in `1.3.71` or `1.4.0`.

charleskorn:
Nice - are you still planning to include this in an upcoming release?

raniejade:
Yeah, `2.1` most likely. Going to make a feature branch for it. I'm planning to reuse the compiler extension you wrote for the JS implementation and maybe jvm as well.

charleskorn:
Awesome, looking forward to it :)

PR #869, opened 2020-03-16
Fix "Test framework quit unexpectedly" error while running tests in IDE

technoir42:
When running a single test in IDE, versions 2.0.6-2.0.10 of Spek fail to discover any tests because any `testDirs` are ignored by classgraph.
@raniejade

### DIFF
--- a/spek-runtime/src/jvmMain/kotlin/org/spekframework/spek2/runtime/JvmDiscoveryContextFactory.kt
+++ b/spek-runtime/src/jvmMain/kotlin/org/spekframework/spek2/runtime/JvmDiscoveryContextFactory.kt
@@ -46,7 +46,7 @@ object JvmDiscoveryContextFactory {
             .enableClassInfo()
 
         if (testDirs.isNotEmpty()) {
-            cg.overrideClasspath(System.getProperty("java.class.path"), testDirs)
+            cg.overrideClasspath(System.getProperty("java.class.path"), *testDirs.toTypedArray())
         }
 
         return cg.scan().use {


raniejade:
@technoir42 Thanks for the PR! Never experienced the problem you are fixing so I'm curious, how did you discover the problem? Any steps I can do to replicate the issue?

technoir42:
I just run a single test using the gutter icon and get "Test framework quit unexpectedly" error every time. Everyone else in my team seems to have the same issue.
The test has to be in an Android module. FWIW I use Open JDK 8, AS 3.6.1, AGP 3.6.1 and Spek plugin 2.0.10.

raniejade:
Ahh, interesting.. I'll have to try it out, thanks! In the meantime, you can build the plugin from source from the `2.0.10` branch (just 🍒 ⛏  your changes). The `2.x` branch contains code that is for the `2.1.0` release.

technoir42:
I found out that this is related to the version of ClassGraph. We use a slightly newer version that Spek's default - 4.8.58 instead of 4.8.37.
I can reproduce the bug on Android sample project by adding:
```
testImplementation "io.github.classgraph:classgraph:4.8.58"
```

raniejade:
Thanks for that @technoir42! I'll also consider upgrading the version of classgraph spek is using.

raniejade:
Backported this to the `2.0.x` branch:https://github.com/spekframework/spek/tree/2.0.x. I'll trigger a release for `2.0.11` this coming week.

PR #903, opened 2020-07-25
Distinguish user land CancellationException from runtime ones

raniejade:
This replaces `withTimeout` call in the executor with `withTimeoutOrNull`.

### DIFF
--- a/spek-kotlin-compiler-plugin-native/src/main/kotlin/org/spekframework/spek2/kotlin/SpekExtension.kt
+++ b/spek-kotlin-compiler-plugin-native/src/main/kotlin/org/spekframework/spek2/kotlin/SpekExtension.kt
@@ -11,8 +11,6 @@ import org.jetbrains.kotlin.descriptors.findClassAcrossModuleDependencies
 import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.builders.*
 import org.jetbrains.kotlin.ir.builders.declarations.addField
-import org.jetbrains.kotlin.ir.builders.declarations.addProperty
-import org.jetbrains.kotlin.ir.builders.declarations.buildFun
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.declarations.impl.IrFunctionImpl
 import org.jetbrains.kotlin.ir.descriptors.WrappedSimpleFunctionDescriptor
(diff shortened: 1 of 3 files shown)


spek-kotlin-compiler-plugin-native/src/main/kotlin/org/spekframework/spek2/kotlin/SpekExtension.kt
```diff
@@ -156,7 +153,9 @@ private class SpekCollector(
     }
 
     private val IrClass.isSpek: Boolean
-        get() = superTypes.any { it.classifierOrFail.descriptor.fqNameSafe.asString() == spekClassName }
+        get() = superTypes.any {
+            it.getClass()?.fqNameForIrSerialization?.asString() == spekClassName
```
raniejade, line 157: not related to this PR, removes the deprecation warning for `it.classifierOrFail.descriptor`.

spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/Executor.kt
```diff
@@ -70,26 +68,38 @@ class Executor {
...
+                            val job = async {
                                 scope.before()
                                 scope.invokeBeforeTestFixtures()
                                 scope.execute()
                             }
 
                             if (scope.timeout == 0L) {
                                 try {
-                                    job.join()
+                                    job.await()
```
raniejade, line 79: `await` allows us to capture the actual exception thrown compared to `launch` which will propagate the exception to the parent context.

spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/Executor.kt
```diff
@@ -108,6 +118,15 @@ class Executor {
         }
     }
 
+    private sealed class TimedExecutionResult {
```
raniejade, line 121: Need to distinguish:
- `null` case, which means test has timed out.
- success case, where the test has succeeded.
- failure case, where the test has failed.

spek-runtime/src/commonMain/kotlin/org/spekframework/spek2/runtime/Executor.kt
```diff
@@ -108,6 +118,15 @@ class Executor {
...
+        object Success : TimedExecutionResult()
+        class Failed(val exception: Throwable) : TimedExecutionResult()
+    }
+
+    private class TestScopeTimeoutException(scopeImpl: TestScopeImpl) : Throwable(
```
raniejade, line 126: For better debugging.

PR #905, opened 2020-07-30
[jvm] Scan classes in parallel

raniejade:
//cc @arturbosch

I'll merge this to the `2.0.x` branch which will trigger a CI build and publish an artifact to: https://bintray.com/spekframework/spek-dev/

### DIFF
--- a/spek-runtime/src/jvmMain/kotlin/org/spekframework/spek2/runtime/JvmDiscoveryContextFactory.kt
+++ b/spek-runtime/src/jvmMain/kotlin/org/spekframework/spek2/runtime/JvmDiscoveryContextFactory.kt
@@ -12,6 +12,10 @@ import kotlin.reflect.full.primaryConstructor
 import kotlin.streams.toList
 
 object JvmDiscoveryContextFactory {
+    private val scanParallelism by lazy {
+        Runtime.getRuntime().availableProcessors()
+    }
+
     private val defaultInstanceFactory = object : InstanceFactory {
         override fun <T : Spek> create(spek: KClass<T>): T {
             return spek.objectInstance ?: spek.constructors.first { it.parameters.isEmpty() }
@@ -44,13 +48,15 @@ object JvmDiscoveryContextFactory {
     private fun scanClasses(testDirs: List<String>): List<KClass<out Spek>> {
         val cg = ClassGraph()
             .enableClassInfo()
+            // any jar on the classpath won't be scanned
+            .disableJarScanning()
 
         if (testDirs.isNotEmpty()) {
             cg.overrideClasspath(System.getProperty("java.class.path"), *testDirs.toTypedArray())
         }
 
-        return cg.scan().use {
-            it.getSubclasses(Spek::class.qualifiedName!!).stream()
+        return cg.scan(scanParallelism).use { scanResult ->
+            scanResult.getSubclasses(Spek::class.qualifiedName!!).stream()
                 .map { it.loadClass() as Class<out Spek> }
                 .filter { !it.isAnonymousClass }
                 .map { it.kotlin }


spek-runtime/src/jvmMain/kotlin/org/spekframework/spek2/runtime/JvmDiscoveryContextFactory.kt
```diff
@@ -44,13 +48,15 @@ object JvmDiscoveryContextFactory {
     private fun scanClasses(testDirs: List<String>): List<KClass<out Spek>> {
         val cg = ClassGraph()
             .enableClassInfo()
+            // any jar on the classpath won't be scanned
+            .disableJarScanning()
```
arturbosch, line 52: This could be huge! Thanks

arturbosch (approved):

raniejade:
@arturbosch new build is out https://bintray.com/spekframework/spek-dev/spek2/2.0.13-alpha.0.1%2Bcecb731 can you try it out?

arturbosch:
> @arturbosch new build is out https://bintray.com/spekframework/spek-dev/spek2/2.0.13-alpha.0.1%2Bcecb731 can you try it out?

I've updated https://github.com/detekt/detekt/issues/2902#issuecomment-666259022

This patch decreases the discovery time by 1 second :tada: . Jupiter still is 1 second faster though.
I will now run build scans for the whole build and keep you updated.

raniejade:
Ohh, that's good news! I'll see if I can configure classgraph further to scan less.

arturbosch:
Our full test time without the gradle plugin decreased by ~ 20s from 2m8s to 1m47s!

Is Spek using a lot of concurrency by default? If yes, maybe this could lead to threads starving when used together with `gradle --parallel`. I know that Jupiter uses opt-in for concurrency.

I see 10seconds discovery time for very simple modules:

With snapshot:
2020-07-30-131741_1191x156_scrot

With 2.0.12:
2020-07-30-131857_1182x66_scrot

spek-runtime/src/jvmMain/kotlin/org/spekframework/spek2/runtime/JvmDiscoveryContextFactory.kt
```diff
@@ -12,6 +12,10 @@ import kotlin.reflect.full.primaryConstructor
 import kotlin.streams.toList
 
 object JvmDiscoveryContextFactory {
+    private val scanParallelism by lazy {
+        Runtime.getRuntime().availableProcessors()
```
arturbosch, line 16: Lazy shouldn't be necessary here?

If I use `gradle --parallel` with 8 cores this will start 8*8 threads to scan the classpath?

raniejade: I was about to link this, and yeah it probably will. I'm still thinking of a way to make this configurable.

raniejade: Maybe via a jvm property? https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:jvmArgs WDYT?

raniejade: It sucks that you can't pass configs from gradle to JUnit Platform test engines.

arturbosch: I think this should be sequential by default and be configurable like you did with the `timeout` property:
`systemProperty("SPEK_TIMEOUT", 0) // disable test timeout`.
Junit5 itself wants to be configured for concurrency with system properties: https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle-config-params

raniejade: Ahh that makes sense - I don't have time now, but I'll look into it in the weekend. Thanks!

raniejade:
I had a closer look, `Classgraph` is already scanning in parallel by default - so adding `disableJarScanning()` probably made the biggest difference.

PR #911, opened 2020-08-06
Make parallel execution configurable (by default it is sequential)

raniejade:


gradle.properties
```diff
@@ -1,2 +1 @@
-kotlin.mpp.enableGranularSourceSetsMetadata=true
```
raniejade, line 1: this is causing issues with composite builds!

PR #955, opened 2021-02-20
Fix docs about CachingMode

sh0n0:
`CachingMode.GROUP` was replaced with `CachingMode.EACH_GROUP`, but remains in docs.


raniejade:
Thank you for fixing this! 🎉

Issue #957, opened 2021-02-28
Incorrect example in a doc section "core-concepts"


ultraon:
Hello, I've found that in a doc section "core-concepts" there is an incorrect example with Kotlin `lateinit var added: Boolean`, link is https://www.spekframework.org/core-concepts/ . Kotlin doesn't support `lateinit` feature for simple types. To fix it you just need to do the following: `var added = false`

raniejade:
Resolved at 14af51d

PR #958, opened 2021-02-28
Fix incorrect example in the "core-concepts" section

ultraon:
This PR provides a fix for the issue https://github.com/spekframework/spek/issues/957


raniejade:
Thanks for fixing!

PR #960, opened 2021-04-02
[2.x] Add support for IJ 2021.1

sh-soltanpour:


snijsure:
@sh-soltanpour Do you need to update spek-ide-plugin-android-studio/build.gradle.kts as well?

More importantly how did you pickup ```"org.jetbrains.kotlin:211-1.4.21-release-IJ6556.4``` this version from various plugins that Intellij publishes.

Also for Android plugin will have to lookup here?

https://plugins.jetbrains.com/docs/intellij/android-studio.html#introduction

raniejade:
@sh-soltanpour you might need to update the kotlin plugin version as well. Have a look here https://plugins.jetbrains.com/plugin/6954-kotlin/versions for the version that is compatible with 2021.1

@snijsure android studio is usually two major versions behind the mainline IJ releases, so no need to do anything for the AS plugin.

sh-soltanpour:
@raniejade No, I think it's not the problem because it's using the latest Kotlin version. 
Could you please take a look at the error now? I've used the latest version published here:
https://confluence.jetbrains.com/display/IDEADEV/IDEA+2021.1+latest+builds

But it says that the new version is not found.

raniejade:
@sh-soltanpour ahh, 211 builds are not in the release repos yet https://www.jetbrains.com/intellij-repository/releases/. They are at the snapshot repos: https://www.jetbrains.com/intellij-repository/snapshots/. You can use `211-EAP-SNAPSHOT` for now.

sh-soltanpour:
@raniejade Done. Thanks

raniejade:
Awesome, can you do the same for the 2.0.x line? There should a branch with the same name :) There are some big changes in the 2.x branch compared to the 2.0.x one. Thanks again @sh-soltanpour

sh-soltanpour:
@raniejade Sure, I've opened another PR for 2.0.x

raniejade (approved):

PR #979, opened 2021-08-13
[2.0.x] Add support for IJ 2021.2

raniejade:
cc @eacheble.

I had to do several things to make it work, but mostly because they are needed to upgrade to Kotlin 1.4.

### DIFF
--- a/integration-test/build.gradle.kts
+++ b/integration-test/build.gradle.kts
@@ -139,7 +139,5 @@ tasks {
 configurations.forEach {
     it.resolutionStrategy.dependencySubstitution {
         substitute(module("org.spekframework.spek2:spek-runtime")).with(project(":spek-runtime"))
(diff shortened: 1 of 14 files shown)


integration-test/build.gradle.kts
```diff
@@ -139,7 +139,5 @@ tasks {
 configurations.forEach {
     it.resolutionStrategy.dependencySubstitution {
         substitute(module("org.spekframework.spek2:spek-runtime")).with(project(":spek-runtime"))
-        substitute(module("org.spekframework.spek2:spek-kotlin-compiler-plugin-jvm")).with(project(":spek-kotlin-compiler-plugin-jvm"))
```
raniejade, line 142: Dropping KN support in this branch, will delete it afterwards.

echeble:
thank you @raniejade for making this work! 🙇

PR #988, opened 2021-12-09
[2.x] Add support for IJ 2021.3

jurriaan:
Upgrade plugin's supporting versions following the release of `IDEA 2021.3`

Based on https://github.com/spekframework/spek/pull/977

### DIFF
--- a/spek-ide-plugin-intellij-idea/build.gradle.kts
+++ b/spek-ide-plugin-intellij-idea/build.gradle.kts
@@ -58,6 +58,7 @@ val buildMatrix = mapOf(
 )
 
 val sdkVersion = project.properties["ij.version"] ?: "IJ213"
+
 val settings = checkNotNull(buildMatrix[sdkVersion])
 
 intellij {
--- a/spek-ide-plugin-intellij-idea/build.gradle.kts
+++ b/spek-ide-plugin-intellij-idea/build.gradle.kts
@@ -49,11 +49,11 @@ val buildMatrix = mapOf(
         arrayOf("java", "org.jetbrains.kotlin:211-1.4.21-release-IJ6693.43")
     ),
     "IJ213" to ij.BuildConfig(
-        "213.5744.223",
+        "213.6461.79",
         "IJ2021.3",
         "IJ183",
         ij.VersionRange("213.1", "213.*"),
-        arrayOf("java", "org.jetbrains.kotlin:211-1.4.32-release-IJ6693.72")
+        arrayOf("java", "org.jetbrains.kotlin:213-1.6.10-release-944-IJ6461.79")
     )
 )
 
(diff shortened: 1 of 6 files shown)


spek-ide-plugin-intellij-idea/build.gradle.kts
```diff
@@ -41,16 +41,23 @@ val buildMatrix = mapOf(
...
         "IJ183",
         ij.VersionRange("212.1", "212.*"),
         arrayOf("java", "org.jetbrains.kotlin:211-1.4.21-release-IJ6693.43")
+    ),
+    "IJ213" to ij.BuildConfig(
+        "213.5744.223",
+        "IJ2021.3",
+        "IJ183",
+        ij.VersionRange("213.1", "213.*"),
+        arrayOf("java", "org.jetbrains.kotlin:211-1.4.32-release-IJ6693.72")
```
jurriaan: Just noticed this isn't the correct version, looking for the correct version number

lizell: `213-1.6.10-release-923-IJ5744.223` maybe?

raniejade: `org.jetbrains.kotlin:211-1.4.21-release-IJ6693.43` should still work. Apologies for the delay, I barely have time to work on personal projects nowadays. If you know someone who wants to take over the project, please let me know!

joelhandwell: @hhariri @jonnyzzz @mhshams @orangy @lkogler any idea for taking over the project (possibly by JetBrains)?

jggomez:
Please, updated to latest version

CriPstian:
Hello :wave: Can we merge this PR? It looks like everything should be fine, or am I missing something? Thanks for this BTW :+1: 
@raniejade is it still up to you to move this forward? How could we help out?

jurriaan:
@CriPstian Unfortunately the PR in the current state is broken. 

I can't get it to compile properly, even with the version constraint here (https://github.com/spekframework/spek/pull/988#discussion_r771910127. It needs to use a later version of the `org.jetbrains.kotlin` plugin. since that plugin is not compatible with the latest intellij anymore.

CriPstian:
@jurriaan The failure might be related to `Jetbrains` moving away from `bintray`.
I knew about bintray being deprecated so now I can see that this build always fails with the same 502 bad gateway.
I believe there should be another task to move to the new repositories, and only then we can move forward with this PR :cry: 
Hope this helps in any way.

jurriaan:
@CriPstian That issue was fixable by removing the bintray repos from the gradle files (it looks like those dependencies are now available on maven), the issue I'm having now is more around compilation of the project with a newer version of `org.jetbrains.kotlin`, version 1.6.x instead of 1.4.x

To me building a zip file that I can manually install in IntelliJ would be very helpful already, but indeed, the spek artifacts should also be moved to a different repository.

CriPstian:
@jurriaan There is also the need to upgrade `gradle` version as it comes with it's own bundled `kotlin` version. Also, tried to upgrade throughout the project where `kotlin` was mentioned.
Also, I did it locally on your branch and had difficulties bringing the `spek-ide-plugin-intellij-idea-base` dependency on the classpath :(
Hope this helps

raniejade:
Closing this in favor of #991. Thanks all for the contribution!

